PR #18, opened 2021-08-30
Make Song Time Bar Optional

BiliousData:
This is pretty self explanatory, it adds an option to hide the Time Bar that shows how much time is left, and is off by default.
Not really much reason to it. I just thought that it should be optional, without disabling the entire HUD.


ShadowMario:
Merged, thanks for this.

PR #38, opened 2021-09-08
Several Tweaks & Changes!

ghost:


Yoshubs:
Made some tweaks to things such as accuracy and note hits. Have yet to revise the input system but you'll probably get to it so I don't wanna take that from you. 

Ghost misses should affect accuracy less. I would nullify that completely but that would take a rewrite of how accuracy works (its score based not exactly judgement based)

Added some small preference changes

Changed hit timings to be double backed (equal on both late and early ends)

Fixed a bug with a lag spike in the options menu

(I was gonna do this over vacation but I had no access to a computer until now)

Yoshubs:
some advice, might wanna make accuracy judgement based just in case anyone decides they want to add custom judgements
what I did is I set a percentage for each judgement to give back (which affected the health they gave back too)

Yoshubs:
most of these are really suggestions you dont have to add these if you dont want to

source/PlayState.hx
```diff
@@ -1836,11 +1839,7 @@ class PlayState extends MusicBeatState
 
 		super.update(elapsed); //TEST
 
-		if(ratingString == '?') {
-			scoreTxt.text = 'Score: ' + songScore + ' | Misses: ' + songMisses + ' | Rating: ' + ratingString;
-		} else {
-			scoreTxt.text = 'Score: ' + songScore + ' | Misses: ' + songMisses + ' | Rating: ' + ratingString + ' (' + Math.floor(ratingPercent * 100) + '%)';
-		}
+		scoreTxt.text = 'Score: ' + songScore + ' | Misses: ' + songMisses + ' | Rating: ' + ratingString + ((ratingString == '?') ? '' : ' (' + Math.floor(ratingPercent * 100) + '%)');
```
Steviegt6, line 1842: Bit of an excessive ternary, don't you think?

Steviegt6: Readability > one-liners IMO.

Yoshubs: totally understood, personally I just like my streamlined code lol 

a better solution were to have a more readable string outside that's edited, then add that to scoreTxt.text, because that's the best of both worlds I think

Steviegt6: I feel 'ya, I use a lot of one-liner stuff for personal projects, but I try to lessen them when I'm making stuff I want people to contribute to. :p
Having some sort of outside string (or a function with parameters you'd pass in) may be better (and promotes arguably less hardcode too).

PR #42, opened 2021-09-09
Turn off anti aliasing of every added sprite instead of having if conditions for every sprite

coolfren:
we do a little trolling


ShadowMario:
setProperty shouldn't be used on every frame, it's not a very optimized function.

PR #43, opened 2021-09-09
take two: disable anti aliasing for every object if globalAntialiasing is disabled

coolfren:
now overrides add() instead of disabling it every frame


ShadowMario:
Doesn't seem to work properly as i've tested

PR #44, opened 2021-09-09
Dialogue bug fixes + New option in Pause Menu

Spel0:
1) Made Dialogue continue on accept keys only

2) Added "Restart Song with Cutscene" option in Pause Menu(Basically repeats whole cutscene sequence instead of skipping straight to song like normal Restart does)

3) Made auto line breaker for dialogue, so it wouldn't go off screen if dialogue is too long, but if it's even longer than 3 lines can handle then it would obviously still clip out of dialogue box bounds (Hello Neo 3.0 :trollface: )

diag

first

second

third


ShadowMario:
I will implement these with the new (not yet released) dialogue changes, thanks for it

Spel0:
Did you add second one in not released version btw? Seems like a good feature in case mod has a cutscene or dialogue and u accidentally skip it

ShadowMario:
I can't approve this now as it seems to be in a different branch, which causes a conflict when i tried to remove your Pause menu changes.
You will still be creditted when i add the extra credits menu tho

PR #85, opened 2021-09-19
Update ClientPrefs.hx

ghost:
Changed The Default Keybinds To DFJK


ShadowMario:
The default keys will continue being WASD + Arrow Keys, you can change them in game if you want to.

PR #94, opened 2021-09-19
Discord Credits RPC Update

ghost:
Forgot to Add It


ShadowMario:
It's correct as it is right now

PR #96, opened 2021-09-20
(Hopefully) Semi-Improved Input System

ghost:
I've attempted to somewhat improve the input system, hopefully it's noticeable and better, there's only one real downside and that's the fact that ghost tapping doesn't work correctly as of right now, at least I think it doesn't, because I'm not missing a lot when I probably should be. Nonetheless, this is open to criticism, changes and edits, so I'm completely fine with changes being made to this as I probably screwed some things up (I'm not very smart lmao)

Hope it's useful!


ShadowMario:
Well, looks better than the current one, i see no reason to not approve it but i will wait a while before doing so (Just in case someone is against thos change)

PR #114, opened 2021-09-23
Update Project.xml

ghost:
This will update Psych Engine's location to a new location like it's seen in Kade Engine instead of using the usual route like it's in the original game. This of course will get the progress of the players deleted since it'll be in a new path. Including the window's title to be called Friday Night Funkin': Psych Engine instead of Friday Night Funkin'.

### DIFF
--- a/Project.xml
+++ b/Project.xml
@@ -2,7 +2,7 @@
 <project>
 	<!-- _________________________ Application Settings _________________________ -->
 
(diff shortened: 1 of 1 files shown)


ShadowMario:
Dang i kinda just didn't wanted to kill saved datas now but alright lol

Steviegt6:
> Dang i kinda just didn't wanted to kill saved datas now but alright lol

Shouldn't be too hard to copy over files if it detects any Psych save data in the original install location?

PR #293, opened 2021-10-10
Add instant respawn option

CerBor:
Adds an option that automatically respawns you after you die, without respawn animation.


ShadowMario:
I still have to fix the options code.

PR #316, opened 2021-10-11
make the case 1 of switch (curBeat) get called

nahiassis:
don't merge this if it doesn't work
i tested it and it works
but it was on a mod
so it's not guaranteed


ShadowMario:
What exactly is this for?

PR #459, opened 2021-10-18
beats

nahiassis:
make the logo bump and the gf dance in  the title screen sync with the bpm of a freeplay song


ShadowMario:
I don't want that

PR #554, opened 2021-10-23
'Info Bar Bounces', 'Max Optimization' and Actions workflows

### LINKED ISSUE #219: Mac and Linux builds.
Can you please make builds that are compatible with Mac and Linux operating system’s please?

Octol1ttle:
'Info Bar Bounces': This option is made because the bouncing was too annoying. I also decreased the info bar's size a bit
'Max Optimization': Self-explanatory
Actions workflows: Automatic builds for Windows, macOS and Linux with binaries. You can use these binaries when publishing releases.
Resolves #219
Resolves #501

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,52 @@
+name: Build Linux
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main ]
+  pull_request:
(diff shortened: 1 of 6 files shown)


steviegt6 (changes requested):
What's with all the random new-lines? Aside from the fact that it appears you're partially enforcing your own coding style, it's terribly inconsistent and non-standard.

l1ttleO:
no longer have time to do this, sorry

PR #583, opened 2021-10-24
Dialogue Enhancements

DevScyu:
Changelog:
```diff
+ Allow dialogue files to be executed without source changes per song
+ Allow dialogue background music to be configured in the dialogue json and via the dialogue editor
+ Adds new dialogueEND support to allow dialogue to be played when a song ends
```

I have tested this to be compatible with 0.4.2.

image

Inside of your `mods/data/SONG` folder:
`dialogue.json` for dialogue when song starts
`dialogueEND.json` for dialogue when song ends

Closes #578

### DIFF
--- a/source/DialogueBoxPsych.hx
+++ b/source/DialogueBoxPsych.hx
@@ -41,6 +41,7 @@ typedef DialogueAnimArray = {
 // love u Shubs no homo :flushedh4:
 typedef DialogueFile = {
 	var dialogue:Array<DialogueLine>;
+	var song:Null<String>;
 }
 
 typedef DialogueLine = {
--- a/source/GameOverSubstate.hx
+++ b/source/GameOverSubstate.hx
@@ -82,6 +82,7 @@ class GameOverSubstate extends MusicBeatSubstate
 			FlxG.sound.music.stop();
 			PlayState.deathCounter = 0;
 			PlayState.seenCutscene = false;
+			PlayState.setOnLuas('seenCutscene', false);
 
 			if (PlayState.isStoryMode)
 				MusicBeatState.switchState(new StoryMenuState());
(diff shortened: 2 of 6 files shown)


FuseIsHere813 (commented):
This looks pretty good, but some Lua stuff are unnecessary as Shadow states.

What am i referencing to is the "setOnLuas" part in line 36 and 68 on GameOverSubState.

DevScyu:
> This looks pretty good, but some Lua stuff are unnecessary as Shadow states.
> 
> What am i referencing to is the "setOnLuas" part in line 36 and 68 on GameOverSubState.

When i was looking at the code, if setOnLuas was not called the variable was not updated, so would this mean, if i do not update it back to false that it would stay in the true state for the lua scripts?

FuseIsHere813:
> > This looks pretty good, but some Lua stuff are unnecessary as Shadow states.
> > What am i referencing to is the "setOnLuas" part in line 36 and 68 on GameOverSubState.
> 
> When i was looking at the code, if setOnLuas was not called the variable was not updated, so would this mean, if i do not update it back to false that it would stay in the true state for the lua scripts?

Maybe.

DevScyu:
> Many of these changes for the Lua related stuff are unnecessary, you don't have to setOnLuas when quitting PlayState because Lua scripts stop running there.

@ShadowMario 
When i was looking at the code, if setOnLuas was not called the variable was not updated, so would this mean, if i do not update it back to false that it would stay in the true state for the lua scripts?

GrexpungedYT:
oh oh
resolve them
081B8D92-FD17-436C-A4CE-CB3C2B63CAB3

frantastic24:
> resolve them

My brother in Christ this is 6 months old

frantastic24:
Why are you bringing up conflicts then

BombasticTom:
this is hella old

Gibberditoslumber:
True

PR #691, opened 2021-10-29
who tf wants to get a miss just for a hurt note

nahiassis:


ItsCapp:
me owo

MeowcaTheoRange:
I mean, aren't you supposed to not hit these? If anything, we could probably toggle this using `ClientPrefs.ghostTapping`

```haxe
// source/PlayState.hx, ln 3329
        health -= daNote.missHealth;
        switch(daNote.noteType) {
		case 'Hurt Note':
			if (!ClientPrefs.ghostTapping) {
                                songMisses++;
			        vocals.volume = 0;
			        RecalculateRating();
			}
		default:
			songMisses++;
			vocals.volume = 0;
			RecalculateRating();
	}
```
Could be a better way to code this, if so, lemme know lmao

nintendofan44:
it would be unfair for people without ghost tapping on, becouse people with it on, won't get a miss from pressing it, only the damage, but for people with it off, they would get the miss
so
this one i have is good already

nintendofan44:
but thanks for caring about this
idk how to code that well
i just did what i remembered

ShadowMario:
Me.

PR #700, opened 2021-10-30
Allow Compilation Without MODS_ALLOWED

steviegt6:
This PR fixes issues with misplaces or missing usages of the `sys` and `MODS_ALLOWED` compilation preprocessors in order to allow compilation of the project without `MODS_ALLOWED`.

## Purpose
Generally just to ensure functionality in the future. This also opens the door for more compatibility between different transpiling targets (namely JS, see Remarks as well).

## Remarks
This does not resolve misusage of classes within the `sys` package namespace, and that should be looked at in a different PR.


AToska21:
@ShadowMario 
might want to take a look at this

nintendofan44:
i tested this and
it kinda broke the lua script loading

Steviegt6:
> i tested this and it kinda broke the lua script loading

There's no reason for it not to function, mind elaborating?

Steviegt6:
Alright, well, in an attempt to make sure Lua stays working, I've removed the extra `MODS_ALLOWED` check for `LUA_ALLOWED`, especially since there should never be a case where `MODS_ALLOWED` is false and `LUA_ALLOWED` would be true othrewise.

prokube:
I did my own version of this and it works at least on Mac. then again this is on mac which nobody uses

FuseIsHere813:
Which means if it gets merged, There might be Mac support.

I approve this, even the bugfixes are great.

FuseIsHere813 (approved):

ShadowMario:
what is the `#if sys` on ChartingState for

prokube:
> Which means if it gets merged, There might be Mac support.

Mac is already technically supported. My fixes just worked on Mac.

Steviegt6:
> what is the `#if sys` on ChartingState for

`sys.FileSystem` isn't important if `MODS_ALLOWED` isn't true, which prevents compilation.

osc-omb:
> I did my own version of this and it works at least on Mac. then again this is on mac which nobody uses

That's because Mac doesn't read Lua.

prokube:
> > I did my own version of this and it works at least on Mac. then again this is on mac which nobody uses
> 
> That's because Mac doesn't read Lua.

i know it doesnt

Steviegt6:
> > I did my own version of this and it works at least on Mac. then again this is on mac which nobody uses
> 
> That's because Mac doesn't read Lua.

The issue is that the library being used doesn't have Mac support (for some God-forsaken reason). Any platform can "read" Lua.

MayoALT:
> > > I did my own version of this and it works at least on Mac. then again this is on mac which nobody uses
> > 
> > 
> > That's because Mac doesn't read Lua.
> 
> The issue is that the library being used doesn't have Mac support (for some God-forsaken reason). Any platform can "read" Lua.

[its currently being fixed from what i know](https://github.com/AndreiRudenko/linc_luajit/issues/10)

PR #755, opened 2021-11-01
make supported versions list

ssdrive-flaktra:
just saying these versions should be the ones most compatible.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,11 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | ✅ Fully Supported |
+| 0.4.x   | :x:                |
+| 4.0.x   | ✔️ Not Fully Supported |
+| < 0.3   | :x:                |
+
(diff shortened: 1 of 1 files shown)


MeowcaTheoRange:
You kinda, uh, *forgot to change the version list*. We're currently on 0.4.2, and according to the list, it's not fully supported, which is kind of silly.

AToska21:
> You kinda, uh, _forgot to change the version list_. We're currently on 0.4.2, and according to the list, it's not fully supported, which is kind of silly.

oops

AToska21:
should be fixed now

Steviegt6:
This is pretty useless. It just completes a generic GitHub repository security task they always ask you to do (which doesn't accomplish anything seeing as this program offers no security in the first place).

Steviegt6:
> This is pretty useless. It just completes a generic GitHub repository security task they always ask you to do (which doesn't accomplish anything seeing as this program offers no security in the first place).

Furthermore, supported versions aren't relevant since different mods will be based on different versions of the engine. This doesn't aid in anything worthwhile.

PR #830, opened 2021-11-05
idk like enable cam zooming if bf hits note?

EyeDaleHim:
why doesn't this exist for bf when it does for dad?


ShadowMario:
Dunno why this works like this on vanilla but yeah i will keep it the same

PR #1053, opened 2021-11-15
remember my god damn volume already

arthu-rguo:
every time i start up friday night funkin' i have to frantically spam the "-" so i don't get my eardrums blasted out by the default max volume. flixel actually saves your volume by default. i have no idea why no one seems to want to load it.


ShadowMario:
https://lib.haxe.org/p/GameJoltAPI
No idea how to mess with this though, good luck.

PR #1086, opened 2021-11-17
much better isn't it

anonybean:
image


ShadowMario:
No, in fact, it's not.
Read this: https://github.com/ShadowMario/FNF-PsychEngine/issues/620

PR #1130, opened 2021-11-18
Add things to lua that would just make me happier lol

KlavierGayming:
Adds an hscript function, which will execute Haxe code when the function is ran. There's some classes defined already.
Adds a getRandomInt and getRandomBool because math.random SUCKS.

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -186,6 +186,22 @@ class FunkinLua {
 			}
 			return Reflect.setProperty(lePlayState, variable, value);
 		});
+		Lua_helper.add_callback(lua, "callMethod", function(method:String, args:Array<Dynamic>) {
+			var pipiss:Array<String> = method.split('.');
+			if (pipiss.length > 1) {
+				var pispiss2:Dynamic = null;
+				if (lePlayState.modchartSprites.exists(pipiss[0]) {
+					pipiss2 = lePlayState.modchartSprites.get(killMe[0]); 	
+				} else {
+					pipiss2 = Reflect.getProperty(lePlayState, pipiss[0]);
+				}
+				for (i in 1...pipiss.length -1 {
+			     		pipiss2 = Reflect.getProperty(coverMeInPiss, pipiss[i]);
+				}
+				return Reflect.callMethod(pipiss2, pipiss[pipiss.length-1], args);
+			}
+			return Reflect.callMethod(lePlayState, method, args);
+		});
 		Lua_helper.add_callback(lua, "getPropertyFromGroup", function(obj:String, index:Int, variable:Dynamic) {
 			if(Std.isOfType(Reflect.getProperty(lePlayState, obj), FlxTypedGroup)) {
 				return Reflect.getProperty(Reflect.getProperty(lePlayState, obj).members[index], variable);
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -198,9 +198,9 @@ class FunkinLua {
 				for (i in 1...pipiss.length -1 {
 			     		pipiss2 = Reflect.getProperty(coverMeInPiss, pipiss[i]);
 				}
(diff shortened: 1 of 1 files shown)


THOBY555:
Just Do Notes Tweens And You Will Make Friday Night Fever XD

KlavierGayming:
what?
this adds some actually i think useful stuff. but idk.

nintendofan44:
would be nice if you explained how those work

KlavierGayming:
getRandomInt will return a number between min and max and getRandomBool Will return a bool depending on the Chance you put in to get a true

KlavierGayming:
hscript() Just runs haxe code that you put inside of the code argument

KlavierGayming:
call Method calls a method

nintendofan44:
> hscript() Just runs haxe code that you put inside of the code argument

like bopeebo rumble?
the modchart for that is on a .hx file

KlavierGayming:
Kind of. you'll need to make loads of modificiations but it will work.

Steviegt6:
> > hscript() Just runs haxe code that you put inside of the code argument
> 
> like bopeebo rumble? the modchart for that is on a .hx file

I believe 4mbr0s3 uses Haxe hotloading, so basically yes. Last I checked, his engine provides an actual modding API and will load separate modules, basically abstracting stuff away and adding hooks for maximum compatibility between mods.

ShadowMario:
> getRandomInt will return a number between min and max and getRandomBool Will return a bool depending on the Chance you put in to get a true

This already exists, it's called math.random

KlavierGayming:
I literally said in the pr description
"Because math.random SUCKS."
It can only generate a new random number every second so you would need to do something Like
'''lua
math.random and math.random and math.random and math.random and math.random and math.random and math.random(1, 2)
'''

KlavierGayming:
i like how you dont even look at the stuff i added, you just see that i added a random number generator and you close it.
We have pretty much no way to do randomBool thats other than 50% on lua!
and we cant call external functions! i added both those things.

KlavierGayming:
ok i got a bit mad there, just gonna make the random int generator in lua...

KlavierGayming:
i made it, if anyone else doesnt like math.random here it is..
contains float, int and bool functions

```lua
local MODULUS = getPropertyFromClass('FlxMath', 'MAX_VALUE_INT');
local internalSeed = 1;
local MULTIPLIER = 48271.0

function generate()
    return internalSeed = (internalSeed * MULTIPLIER) % MODULUS
end

function int(Min, Max)
    if Min == 0 and Max == MODULUS then
        return math.floor(generate());
    elseif Min == Max then
        return Min;
    else
        if Min > Max then
            Min = Min + Max
            Max = Min - Max
            Min = Min - Max
        end

        local result = 0
        
        result = math.floor(Min + generate() / MODULUS * (Max - Min + 1))

        return result;
    end
end

function float(Min, Max)
    if Min == 0 and Max == 1 then
        return generate() / MODULUS;
    elseif Min == Max then
        return Min;
    else
        if Min > Max then
            Min = Min + Max
            Max = Min - Max
            Min = Min - Max
        end

        local result = 0
        
        result = Min + generate() / MODULUS * (Max - Min + 1)

        return result;
    end
end

function bool(chance)
    return float(0, 100) < chance;
end
```

KlavierGayming:
i tried copying FlxRandom's algorithm for randomizing

BobbyDrawz:
> hscript() Just runs haxe code that you put inside of the code argument

uhh you know there's runHaxeCode()?

Rudyrue:
> uhh you know there's runHaxeCode()?
this pr is from 2021

BobbyDrawz:
> > uhh you know there's runHaxeCode()?
> 
> this pr is from 2021 runHaxeCode wasn't added until like a year later

ty for info

PR #1289, opened 2021-11-23
Update GameOverSubstate

afeefflox:
Remove From PlayState in GameOverSubstate Setting


ShadowMario:
Why???????
This is needed for the Week 6 custom death screen things

PR #1567, opened 2021-12-01
Stop the errors that people are complaining about (If its still going on)

CrusherNotDrip:
You can use this for now until your actual code is done for ModsMenuState.hx so that way not many duplicates are in issues (I hope this works if it doesnt tell me the problem)

### DIFF
--- a/source/ModsMenuState.hx
+++ b/source/ModsMenuState.hx
@@ -385,15 +385,18 @@ class ModMetadata
 				var stuff = Json.parse(rawJson);
 				if(stuff.name != null && stuff.name.length > 0)
 				{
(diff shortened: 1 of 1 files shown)


CrusherMods:
I may of picked the wrong color sorry about that but you can fix it if you want

Blockyyy:
It works!

ChargerDoinDaCode:
Still isn't workin.

DaBanana64:
The error has been fixed on the latest commit.

CrusherMods:
> The error has been fixed on the latest commit.

It has??? alright im shutting it down

PR #1582, opened 2021-12-02
Update CoolUtil.hx

liew757:


HiroOfc:
what

franjavierr:
We don't need more difficulties to the base game, add it to your mod or something

Doritojoey:
every song would crash if there wasnt a chart made this would just be a whole other load of work for psych 5

nintendofan44:
@liew757 yknow there is a thing for loading custom diffs in the freeplaystate
and it works on the storymode menu too
even if you need to add the texture on a folder
still, there already is something that you can add more diffs
and you can like add 2 more to a song and 5 more to another and that would work fine

HiroOfc:
yea, they're right
but something i think nobody noted: **he forgot a comma at the code**

SrtHero278:
Welp, we know what request is being closed.

DaBanana64:
You can literally add custom difficulties on the latest commit. (Only works in Freeplay, I think it's still being worked on) Close this please.

Steviegt6:
@ShadowMario

ShadowMario:
Stupid + Outdated + Ratio + Take this L

PR #1610, opened 2021-12-02
Swap section note datas

tposejank:
What was added:
### Two buttons to swap notes on both sides of the section.
Each button will invert notedata of a note in the left side (or right side) of the current section. Example:
I have a left note in the left side of the current section. I will click on the "Invert notes (left side)" button and it will become a right note.
Same with the right side.
It does not affect event notes.

Useful if you're charting a song with a lot of random patterns. I don't really know lol.


bbpanzu:
sounds nice. i'll consider it

PR #1878, opened 2021-12-09
Fixed an error in StoryMenuState

flashintv:
`image is FlxGraphic` changed to `Std.is(image, FlxGraphic)`


flashintv:
Creating this pull request, because I seen a lot of people coming and posting about: `Expected )`

flashintv:
> change `Std.is` to `Std.isOfType`

both of them are the same, but the other one is just depricated but sure

ShadowMario:
Update your Haxe.

flashintv:
> Update your Haxe.
caption

No legacy support

PR #2115, opened 2021-12-14
Allow the use of \n with debugPrint / luaTrace

tposejank:
### What it will do

* Creates various lines from an array split by \n from a string.


arezzg (approved):
Damn this actually cool and useful thing.

PR #2229, opened 2021-12-16
Controller support (not really sure if it gives the right results)

tposejank:
I'm using a PS3 controller, I had this idea, I put a bunch of stuff in a new Menu, and it seems pretty good.
It might not give the right inputs, but It does what it's supposed to.

### DIFF
--- a/source/ClientPrefs.hx
+++ b/source/ClientPrefs.hx
@@ -1,5 +1,6 @@
 package;
 
+import flixel.input.gamepad.FlxGamepadInputID;
 import flixel.FlxG;
 import flixel.util.FlxSave;
 import flixel.input.keyboard.FlxKey;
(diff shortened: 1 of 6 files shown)


cfgexe (approved):
WOw this is cool

Gui-iago:
at least with my xbox series X controller, it doesn't give the rights results, for example, if I move the left stick to any direction, it can either bind it as "B", "Y" or "X", when in-game, I need to press those exact buttons (instead of moving the left stick) in order to do said action

cfgexe:
we need to add a thing to options menu where it asks you if you have a xbox or playstation controller and then modifies the results

tposejank:
> we need to add a thing to options menu where it asks you if you have a xbox or playstation controller and then modifies the results
> 

i was planning on doing that but
i thought would someone even care to use this
i mean, like this you would only see it on the menu.

cfgexe:
we could use save data and determine if this is the first time ran with a controller and ask then?

cfgexe:
bruh

PR #2287, opened 2021-12-18
Fix score adding, accuracy & rating adjustments

Octol1ttle:
currently:
If rating is shit add 350
If rating is bad add 350
If rating is good add 350
If rating is sick add 350
with this PR:
If rating is shit add 50
If rating is bad add 100
If rating is good add 200
If rating is sick add 350
All values here are according to the vanilla game


Gui-iago:
this actually affects the gameay in such a way that getting high accuracy values is complicated thanks to the "Good" rating now being a perfect way to screw your SFC up

what anout  a Accuracy Type option instead? so these values could still be used as Hard Mode for Accuracy and Score Values

l1ttleO:
> this actually affects the gameplay in such a way that getting high accuracy values is complicated thanks to the "Good" rating now being a perfect way to screw your SFC up and at the same time give you 95% Accuracy or less
> 
> what about a Accuracy Type option instead? so these values could still be used as Hard Mode for Accuracy and Score Values, could be something like this [Gui-iago@dc8c96e](https://github.com/Gui-iago/FNF-PsychAmalgam/commit/dc8c96ec9d21d7a83871141573a79b0efa212101) (ignore CoolUtil.hx, I'm just dumb)

@Gui-iago It doesn't affect your accuracy because it's not calculated depending on your score
It's calculated by dividing `totalNotesHit` by `totalNotesPlayed`. `totalNotesHit` is modified depending on your note rating ('Shit' adds 0, 'Bad' adds 0.5, 'Good' adds 0.75, 'Sick' adds 1). `totalNotesPlayed` is all notes that appeared on Boyfriend's lane (including missed ones)

TheLeerName:
so thats why my friend played on my chart ported from KE and got so many points with such low accuracy, i didnt notice that xd

l1ttleO:
github is weird and added like 10 people here because I pulled from remote
Anyway, I pushed a commit that uses the `Perfect!!` rating instead of `?` before a note is played, and adds `0.25` to `totalNotesHit` if your rating is `Shit` instead of `0`

PR #2298, opened 2021-12-18
Fixed Lua Sprites :/

afeefflox:
idk why Added antialiasing for Pixel Background


tposejank:
This isn't needed. You can just do setProperty('blahblah.antialiasing', false)

afeefflox:
oh

PR #2350, opened 2021-12-19
Hide mouse after exiting some editors

TheLeerName:
You forgor 💀 add `FlxG.mouse.visible = false;` after switch to master editor menu lmao


HiroMizuki (approved):
pog

PR #2556, opened 2021-12-23
Some more functions for Luas

tposejank:
map functions and 2 functions for noobs that do not know how to use lua's locals
# what i used to test it lol
lol

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -282,7 +282,31 @@ class FunkinLua {
 			}
 			Reflect.getProperty(getInstance(), obj).remove(Reflect.getProperty(getInstance(), obj)[index]);
 		});
-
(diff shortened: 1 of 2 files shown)


SrtHero278:
for the custom variables, I know you said it for people who don't understand but it's SO EASY. Just do this:
```
local variable = value
```

tposejank:
> for the custom variables, I know you said it for people who don't understand but it's SO EASY. Just do this:
> 
> ```
> local variable = value
> ```

i fucking know

PoleyMagik:
Regarding the "functions for noobs", I don't think this will help at all. If anything this just makes programming in lua infinitely more unintuitive. It was already fine and these additions merely serve to overcomplicate a simple process. The map editing functions also appear to be superfluous to an extent, and it appears your commit contains a good amount of broken code as you yourself have notated.

Overall, appears to be a bad pull request.

PR #2607, opened 2021-12-23
Minor nitpicks and sum changes

Trayfellow:
- Flicker effect when selecting a song in freeplay
- selecting a thing (song/week) in freeplay and story mode now pauses the music
- Menu image size isnt increased anymore in some places like the options menu cause the image doesnt need to scroll up/down there


trayfellow:
oh wait i found out what the shift thing does lel

PR #2638, opened 2021-12-24
Update Discord.hx

MAJigsaw77:


PoleyMagik:
Combine your commits into one pull request next time, but other then that I don't see anything immediately wrong with this.

PR #2656, opened 2021-12-24
Set require to false, prevent running external lua libraries that could cause a security risk

superpowers04:
LuaJIT has ffi, which allows execution of raw C code, this should prevent against importing it


ShadowMario:
Before merging this i have to know, is there something important that is lost from this? Like, not being able to use some actually useful function like idk, `math.floor` for example

superpowers04:
Nope, those functions should not be handled with require. Atleast in normal lua, they're always present

RonHereLMAO:
does this mean i cant no more making people crash their games

PR #2873, opened 2021-12-29
Fixed Cocoa Charts

MasterDirector99:
It fixes all the difficulty charts for the one song called Cocoa!


MasterDirector99:
Don't merge it yet, I had to fix both normal & hard

ShadowMario:
What was the issue with it???

tposejank:
they probably weren't like the chart generating right now. It wasn't even a issue either

MasterDirector99:
Some changes

Normal: Left Right Left Right Up was changed to just Left Left Up
Easy & Hard: Alt Animation enables when Mommy Mearest finished sang

HiroMizuki:
removing the left right left right up on normal is rlly unnecessary cuz it's normal mode, it would be understandable if that chart was at easy

ShadowMario:
No.

PR #2892, opened 2021-12-30
Hey i added a new color to philly lights i think it would be good to merge check it out

blaidev:
Hey i added a new color to philly lights i think it would be good to merge check it out


SrtHero278:
I don't think this will be merged since he doesn't like blammed lights nowadays. (Overused)

HiroMizuki:
definitely no, shadow mario would never merge it

skedgyedgy:
-- this is just a darker shade of a color that's already there

PR #2951, opened 2021-12-31
Audio and Visual Caching Overhaul

ghost:


tposejank (approved):
@ShadowMario @bbpanzu the j

Yoshubs:
Pros:

- Better memory management!
- Best of both worlds
- Returns assets instead of paths when using paths.hx
- Buttery smooth when restarting and going into chart editor
- Doesn't lag much with Lua

Cons:
- Paths.hx no longer returns the actual path unless you're using getPath (very easy to circumvent in most cases)
- not 100% tested with mod support
- Removes the persistent caching option because this does that but without destroying your computer

tposejank:
when this finishes doing the action imma see how my shit pc can handle psych then

gaminbottomtext:
Ok, this is cool.

haya3218 (approved):
Vouch :)

PR #3001, opened 2022-01-01
Artemis integration (i.e. pretty blinking lights for your beep boop game)

skedgyedgy:
hi did you ever want your pretty keyboard lights to react to the stuff that happens in your funny beep boop man simulator rap battle rythm game well now you can!!!

-- added in an integration to artemis rgb which allows implementing rgb effects on basically any keyboard/mouse/whatever that allows that sort of control
-- stuff like the keyboard changing "backgrounds" to match with the level, changing the whole color during blammed, flashing during lightning strikes, fading to match the background colors of menus, etc. etc.
-- lua support
-- cool rainbow effect when you hold a full combo for long enough
-- fully customizable
-- o p t i o n a l
-- allows for custom trigger and values for modability and versatility and whatever

https://github.com/skedgyedgy/Artemis.Plugins.FNF/releases/tag/1.0
https://youtu.be/9A2IIk0EDN4

Screenshot (42)
Screenshot (44)
Screenshot (45)
Screenshot (46)
Screenshot (47)
Screenshot (50)


tposejank (approved):
sick af, don't think shadow mario will approve it though.

skedgyedgy:
-- still a draft because i realized there's like a major improvement i could add to this that'd make it far easier to use for mods

source/ArtemisIntegration.hx
```diff
@@ -0,0 +1,255 @@
...
+import haxe.Http;
+import haxe.io.Path;
+import sys.FileSystem;
+import sys.io.File;
+#end
+
+using StringTools;
+
+class ArtemisIntegration {
+    // public static inline var ArtemisAPIUrlDirectory:String = "%ProgramData%/Artemis"; // WHAT DO YOU MEAN THIS DOESN'T WORK???????
```
ActualMandM, line 15: You have to use `Sys.getEnv("ProgramData")` instead of `%ProgramData%`

skedgyedgy:
-- i'm working on that thing i was talking about here: https://github.com/skedgyedgy/FNF-PsychEngine/tree/profile-control
-- basically allowing mods to include their own profile data and send them over themselves. i still need to flesh this out a lot more plus do some cleanup but it's late, you can take a look at what i've done so far if you'd like

skedgyedgy:
now rather than the user having to install profiles, mods can include artemis profiles "MODFOLDER/artemis/default.json" which will automatically be added to artemis. there is also a vanilla profile ("assets/artemis/fnf-vanilla.json" by default) which can be edited by mod creators if they're actually editing the source code of the engine for their mod instead of a "drag-into-the-mods-folder" type of mod

bbpanzu:
cool idea. bookmarking

skedgyedgy:
-- added a couple more of the things i wanted to add like the controls now changing on the keyboard to match your own custom controls instead of being stuck at WASD and arrow keys, and the mod can now control the colors of the control keys as well. this has also been done in a way which will support custom control such as 5 key or whatever since they work on a dictionary/map rather than just four variables named "left down up right"
-- also 8 custom booleans added to the integration that modders will be able to set that can be used for whatever they want

-- i'm getting close to completion. should i go through and surround all the points where artemis functions are called with "#if sys"? the functions are ignored if sys isn't supported since the "enabled" flag will never be set but idk if that'll still be necissary

skedgyedgy:
-- anyone wanna review this and make sure i didn't do anything majorly bad and awful? i'm still hesitant on taking this out of draft mode

Steviegt6:
After some quick glances, your formatting is a little off (`functionName (parameters)` doesn't line up with the standard `functionName(parameters)` used in this repository, but that's pretty insignificant. The first major issue I see is a hardcoded port number - especially one that is by no means unique (9696). Would be safer to just scan for unoccupied ports and default to some really random port that no one is likely to already use. Other than that, it seems fine, but traces appear to be a little spammy and generally undescriptive.

skedgyedgy:
-- there is no hardcoded port number; there isn't even a server on fnf's side in the first place. that's just a default value i put in there during testing and left in, the first step of the program is to scan the "%ProgramData%/Artemis/webserver.txt" file for the actual address which replaces that value

XtraXD1:
Good idea! I don't have one of these, but I like the idea. Hope that ShadowMario says the same :)
(Sorry for any grammar error lmao)

XtraXD1:
Woo! So cool! What does it do? Recreate the screen, but in the keyboard? By the way, does it also color the custom keybinds? I mean, if you use "QWOP" to play, instead of "ASWD", if it colors QWOP of ASWD.

skedgyedgy:
@XtraXD1 -- thanks! also nope! i put in custom visuals for each of the vanilla stages, plus i'm making some for other mods too. i've done madness vandalization and am working on vs. sonic.exe next.

-- also yeah, the keybinds automatically update to match your own

MrClogsworthYT:
> @XtraXD1 -- thanks! also nope! i put in custom visuals for each of the vanilla stages, plus i'm making some for other mods too. i've done madness vandalization and am working on vs. sonic.exe next.
> 
> -- also yeah, the keybinds automatically update to match your own

why the hell do you you add -- to all your messages
like

-- hi

skedgyedgy:
-- 'cause it's comforting to me

HiroMizuki (approved):
oooo i forgor

HiroMizuki:
ik that maybe will not be merged, but i liked these

SonicProYT:
make it compatible with 0.5.2h pls

Joalor64 (approved):
you need to be in the team this is amazing

ShadowMario:
That's cool and all but this integration could be an issue for mods and isn't really that relevant.
It sounds mostly like something that would exist in a mod instead of an engine itself.

Issue #3025, opened 2022-01-02
I Have Found A Very Annoying Custom Events Bug


CC-J4xon:
### Describe your bug here.

I Use The "Change Character" Event, I Do This:

Value 1: Dad
Value 2: amogus-hell

(The Character Is Not Daddy Dearest, But They Are On The Opponent's Side)
It Changes BF To "amogus-hell" Instead Of The Opponent.

So I Try This:

Value 1: amogus
Value 2: amogus-hell

(The Character's Name Is Amogus (The Opponent (The One I Want To Change)))
It **_STILL_** Changes BF To "amogus-hell"

**HELP**

### Are you modding a build from source or with Lua?

Lua

### What is your build target?

Windows

### Did you edit anything in this build? If so, mention or summarize your changes.

I Added A Couple Of Custom Events And New Characters

Starmapo:
You have to make value 1 lowercase. Like this:
```
Value 1: dad
Value 2: amogus-hell
```

CC-J4xon:
> You have to make value 1 lowercase. Like this:
> 
> ```
> Value 1: dad
> Value 2: amogus-hell
> ```

Thank You :D

PR #3034, opened 2022-01-02
Lowercase and trim event character values

Starmapo:
Formats the 'Alt Idle Animation' and 'Change Character' character type values. Mainly to prevent stuff like #3025, plus other events like 'Hey!' and 'Play Animation' already do it


TheSmoothWarrior (approved):
I was thinking something was up when I tried to port Vs. Void to Psych Engine & had these same issues.
Honestly, this is a change that has been long overdue & should already be implemented by now.

PR #3035, opened 2022-01-02
Profile control

skedgyedgy:
now rather than the user having to install profiles, mods can include artemis profiles "MODFOLDER/artemis/default.json" which will automatically be added to artemis. there is also a default profile which can be edited by mod creators if they're actually editing the source code of the engine instead of a folder mod


skedgyedgy:
mistake, i meant to pull this into my own fork idk why it defaulted here

Issue #3017, opened 2022-01-02
Quicker Offsetting in character editor


BeansCool54:
### What feature do you want to get added on the **base** engine?

The idea is pretty simple. Making it so that when you hold arrow keys it moves automatically and you dont have to mash them.

### To test your sight, and reliability, please select the option of what should NOT be requested.

Stage Editor, 6K+ support, and winning icons.

HiroMizuki:
there's already something at the char editor that moves the character 10x faster if u hold it

BeansCool54:
> there's already something at the char editor that moves the character 10x faster if u hold it

I know. What I was trying to say is that it's automatic and you dont need to press the button every time to make the sprite move. You can just hold the arrow keys and it will move

HiroMizuki:
> > there's already something at the char editor that moves the character 10x faster if u hold it
> 
> I know. What I was trying to say is that it's automatic and you dont need to press the button every time to make the sprite move. You can just hold the arrow keys and it will move

ok
well, maybe that's a good idea

PR #3041, opened 2022-01-02
Made it possible to hold down the Arrow Keys in CharacterEditor

lemz1:
Hi,
I added the feature to hold down the Arrow Keys to move the Character by holding down Control.
I added it since someone mentioned it in issue #3017


feder273:
u are a live saver

aflacc:
Honestly I don't see why this is needed

PR #3138, opened 2022-01-04
Game-Window movement/positioning for lua

lemz1:
Hi I made it possible to move the game-window in lua.
I also made it move to its original x,y,width,height after the song was restarted or exited.

You can move the game-window in Lua by using, for example, this code:
```
function onUpdate()

    songPos = getSongPosition()
    local currentBeat = (songPos/1000)*(bpm/60)
        setProperty('window.x', windowX + math.sin(currentBeat) * 15)--X axis
        setProperty('window.y', windowY + math.cos(currentBeat) * 15)--Y axis
        setProperty('window.width', windowW + math.sin(currentBeat) * 15)--height
        setProperty('window.height', windowH + math.sin(currentBeat) * 15)--height

end
```

[width and height have a huge performance issue when using the game-capture in a video-recording-software, because the window size is being updated in the video-recording-software. Using the screen capture fixes the performance issue.]

### Issue that still needs to be fixed if too bad ###
_the fps display number keeps getting higher when using any width or height movement. The number resets when the game is paused or the song is exited
the fps display number gets a little bit higher when using width or height, but it doesn't keep getting higher._

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -110,6 +110,12 @@ class FunkinLua {
 		set('screenWidth', FlxG.width);
 		set('screenHeight', FlxG.height);
 
+		// Window shit
+		set('windowX', PlayState.instance.window.x);
+		set('windowY', PlayState.instance.window.y);
+		set('windowW', PlayState.instance.window.width);
+		set('windowH', PlayState.instance.window.height);
+
 		// PlayState cringe ass nae nae bullcrap
 		set('curBeat', 0);
 		set('curStep', 0);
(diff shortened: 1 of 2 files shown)


skedgyedgy:
> I also made it move to its original x,y,width,height after the song was restarted or exited.

-- i feel like this part should be left out, that would be more jarring than helpful. the window should just stay where it is when the song ends even if it got moved

bbpanzu:
> the fps display number keeps getting higher when using any width or height movement. The number resets when the game is paused or the song is exited the fps display number gets a little bit higher when using width or height, but it doesn't keep getting higher.

that's a openfl issue. not really an issue anyway

skedgyedgy:
-- i can't seem to get it to work

Itz-Miles:
Have you tested your PR with the most recent release or commit on Psych Engine? I noticed your fork is outdated, and this doesn't seem to work on 0.5.1

PR #3245, opened 2022-01-06
Fix charting waveform

Starmapo:
This fixes the audio waveform not showing up in the chart editor, which was broken after the new caching method was added (returned a sound file instead of a text path).

There is probably a way to get an audio buffer straight from a Sound file instead of having to use a path, but I'm not smart enough for that.


FurryColtin:
hey i am gonna merge this to my psych engine but better github page that ok with you?

Starmapo:
> hey i am gonna merge this to my psych engine but better github page that ok with you?

You can go ahead

PR #3374, opened 2022-01-09
added scaling modes

skedgyedgy:
recently an update was pushed that disabled alt+enter in order to allow for adaptive aspect ratios which, i mean, fair at first but StageScaleMode is buggy, it should not be the default and it really should NOT be prioritized over the ability to toggle fullscreen in-game. i added a screen scaling mode option, changed the way the resolution is applied so it's triggered by pressing enter instead of backing out into the main menu, and i made it so you can use alt+enter again unless the player is in adaptive mode which is the mode where alt+enter would cause problems


skedgyedgy:
wrong branch

skedgyedgy:
-- idk why it just randomly decided to add every commit from a completely different branch i'm sorry

PR #3413, opened 2022-01-10
Update FunkinLua.hx

Zoher098:


skedgyedgy:
-- you really shouldn't just make pull requests without so much as a title or description saying, like, what your pull request is actually doing

skedgyedgy (changes requested):
-- this looks like it's just going to break existing scripts, what is the point of this change?

source/FunkinLua.hx
```diff
@@ -43,8 +43,8 @@ import Discord;
 using StringTools;
 
 class FunkinLua {
-	public static var Function_Stop = 1;
-	public static var Function_Continue = 0;
+	public static var Function_Stop =''Function_Stop'' ;
```
tposejank, line 46: what

source/FunkinLua.hx
```diff
@@ -43,8 +43,8 @@ import Discord;
...
+	public static var Function_Continue = ''Function_Continue'';
```
tposejank, line 47: what

skedgyedgy: -- yeah literally all this will do is create a minor performance drop for something that benefits nobody

source/FunkinLua.hx
XtraXD1: wtf

charlesisfeline (commented):
ugh

skedgyedgy:
-- thank you for the necropost kind stranger

SonicProYT:
im not necroposting, it looks like its gonna break lua scripts

skedgyedgy:
-- it won't break lua scripts, it really won't do much of anything other than create a sliver of a degrade in performance for no benefit
-- also i don't think you know what necroposting is 'cause randomly commenting just "ugh" on a months old thread is definitely that

SonicProYT:
but i said ugh cuz how:

add winning icons = you broke the icon health system

@ShadowMario close this pls

SonicProYT:
and it impacts performance

skedgyedgy:
-- this doesn't add winning icons

Issue #3369, opened 2022-01-09
week editor bug


HiroMizuki:
### Describe your bug here.

there is a bug on the week editor
when u select a storymenu file, it says that it doesn't exist, but this actually exists
![image](https://user-images.githubusercontent.com/72695242/148692522-fecadadb-6cf4-4add-8880-04bb32fd7c89.png)

idk what is this, but it happens

### Are you modding a build from source or with Lua?

Source

### What is your build target?

Windows

### Did you edit anything in this build? If so, mention or summarize your changes.

no, just tried to add a week

HiroMizuki:
discovered the error
i'll try to make a PR

PR #3456, opened 2022-01-11
fix assets files search at week editor

HiroMizuki:
i made a bug issue (#3369) that shows the week editor not loading automatically the week and stage bg image assets
i promised that i would make a PR to fix this, and here it is

actually, the error was that the week and bg images was just loading when u write a ".png" after the file name, but the preset was not with this
so, this PR fixes this, now, u don't need to write ".png" after the file's name

small PR? Yes, but it fixes something


Starmapo (approved):

ghost (approved):

tposejank (approved):

nahiassis (approved):
pog

PR #3608, opened 2022-01-15
Update README.md

fireredfan7:
Using the git version makes sure it’s updated.


Stilic:
yeah the github have a fix for makes lua works on mac, not the haxelib version.

PR #3373, opened 2022-01-09
Softcoded achievements

tposejank:
example of an achievement

```
{
    "name": "Name here",
    "icon": "image name here",
    "description": "Description ",
    "unlocksAfter": "the week it unlocks after",
    "hidden": (is it hidden? true or false)
    "customGoal": (is this goal customizable? true or false)
}
```

for custom goals, when the game checks for achievements it calls function `onCheckForAchievement` with arguments `name`
do return Function_Continue to notify the game it is unlockable
example:

`if getProperty('misses') > 10 and leName == 'lmao_skill_issue' then return Function_Continue end`

on the luas.

PR #3625, opened 2022-01-16
Revert "Softcoded achievements"

tposejank:
Reverts ShadowMario/FNF-PsychEngine#3373 

I change my mind, they are so glitched out, they don't even work please revert it


tposejank:
Despite this not working i will keep making more banger ahits

PR #3684, opened 2022-01-17
The thing

MAJigsaw77:


fnaffan29:
what

tposejank:
?????????????????

jigsaw-4277821:
In html 5 is not used

jigsaw-4277821:
Or js

jigsaw-4277821:
Only on desktop

jigsaw-4277821:
So the reason

PR #3818, opened 2022-01-22
Adding drag and drop for character editor

### LINKED ISSUE #3053: Using mouse on offset editor
### What feature do you want to get added on the **base** engine?

Being able to do drag n drop shit on the character editor as an alternative to using arrow keys. I think kade has this but I dunno

### To test your sight, and reliability, please select the option of what should NOT be requested.

Stage Editor, 6K+ support, and winning icons.

blueDev2:
Since FlxG.mouse.overlap() does not take into consideration of offsets, it could not be used to make drag and drop only occur when you click on the character. 
So Instead, drag and drop occurs at all times when you click on anything and move the mouse, except when the mouse is over a UI box. This is to prevent accidental drag and drop when clicking on a drop-down menu. 
resolves #3053

### DIFF
--- a/source/editors/CharacterEditorState.hx
+++ b/source/editors/CharacterEditorState.hx
@@ -154,7 +154,9 @@ class CharacterEditorState extends MusicBeatState
 			\nArrow Keys - Move Character Offset
 			\nR - Reset Current Offset
 			\nHold Shift to Move 10x faster
-			\n", 12);
+			", 12);
+			/*\nOptionally: Drag and drop to move animation
+			\nShift to drag the entire character*/
 		tipText.cameras = [camHUD];
 		tipText.setFormat(null, 12, FlxColor.WHITE, CENTER, FlxTextBorderStyle.OUTLINE, FlxColor.BLACK);
 		tipText.scrollFactor.set();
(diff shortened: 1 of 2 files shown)


Starmapo (approved):

cherryiguess:
so you can move the camera position and character offsets (or placement) with your mouse? feels more convenient and useful then clicking the same button over and over until its good

blueDev2:
I think there is a bit of confusion.
This changes the position of an animation, not the entire character.
The latter is possible, but not with the current setup.

HiroMizuki (approved):

ImaginationSuperHero528:
> I think there is a bit of confusion. This changes the position of an animation, not the entire character. This is an alternative to the shift and arrows. The latter is possible, but not with the current setup.

For entire character it could just be hold shift

blueDev2:
Advise was taken and implemented.
Though I would have liked to add text about it, but there isn't much space

Lsoissus:
woo, this is pretty cool, i will add it to my version of psych engine made in source code [ill give credit]

blueDev2:
I'm glad somebody has found use in this. I'm not sure if ShadowMario has even noticed this pull request yet, but I think it's at least ok since a lot of other requests have been closed by him.

PR #3921, opened 2022-01-24
Note scaling stuff

lemz1:
Hi,
this just makes the notes be the scale of the strums.
I thought this was needed since (atleast from what I know) you can't make the note's scale update every frame with lua.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -2507,6 +2507,8 @@ class PlayState extends MusicBeatState
 				var strumAlpha:Float = 0;
 				var strumDirection:Float = 0;
 				var strumScroll:Bool = false;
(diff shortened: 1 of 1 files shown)


tposejank:
no

min89xd:
@tposejank  why do you put no in every pull request?

tposejank:
because its useless

Steviegt6:
> because its useless

It'd be more helpful to elaborate on why a pull request is useless, maybe offering some criticism on how it could improve.

MayoALT:
> because its useless

you could argue that your achievement pr was useless because you can already save data in psych engine and recreate the achievement unlocking thing

PR #3983, opened 2022-01-26
Fixing Sustain Notes When The Direction is Changed

bdidk235:
Fixing some stuff, I like changing direction for notes!


HiroMizuki:
maybe u could show some video so people can see what it does

AdrianCD5:
This is a really good fix, I need this to force the opponent to remain on upscroll and the sustains are buggy with the new note system. Thanks :)

bdidk235:
The sustain arrows are not reducing, I'm trying to fix it.

KevinKuntz255:
would really love this pulled since I'm making this

https://user-images.githubusercontent.com/35108759/151361652-f3e1b63e-7a35-4120-955c-f919a99b4ca0.mp4

nintendofan44:
the sus clip isnt on the center anymore if you didnt notice it yet

PR #4006, opened 2022-01-27
Disable os, to prevent running raw commandline shit

superpowers04:


crowplexus (approved):

stilic (approved):
NOT MORE RICKROLL!!!!! (but good for protecting people from any hacker script)

Jotaro-Gaming (approved):
good

MeguminBOT (approved):

tposejank (approved):
yes

steviegt6 (approved):
Looks fine, but I believe this is only really a band-aid fix. Some discussion regarding a plugin permission system may prove useful to allow niche things to still work.

fnaffan29:
rip "create text file at the start of the song" command

fnaffan29:
nice

lunacynne (changes requested):
this change would break the mod im working on since it uses functions like os.clock and os.exit
would it be possible to disable only the potentially malicious functions like os.execute and os.getenv? or is this not possible

like would this work
```
set('os.execute', false);
set('os.getenv', false);
set('os.remove', false);
set('os.rename', false);
set('os.setlocale', false);
set('os.tmpname', false);
```

tposejank:
> this change would break the mod im working on since it uses functions like os.clock and os.exit would it be possible to disable only the potentially malicious functions like os.execute and os.getenv? or is this not possible
> 
> like would this work
> 
> ```
> set('os.execute', false);
> set('os.getenv', false);
> set('os.remove', false);
> set('os.rename', false);
> set('os.setlocale', false);
> set('os.tmpname', false);
> ```

malicious intent

Stilic:
> > this change would break the mod im working on since it uses functions like os.clock and os.exit would it be possible to disable only the potentially malicious functions like os.execute and os.getenv? or is this not possible
> > like would this work
> > ```
> > set('os.execute', false);
> > set('os.getenv', false);
> > set('os.remove', false);
> > set('os.rename', false);
> > set('os.setlocale', false);
> > set('os.tmpname', false);
> > ```
> 
> malicious intent

wdym

superpowers04:
> this change would break the mod im working on since it uses functions like os.clock and os.exit would it be possible to disable only the potentially malicious functions like os.execute and os.getenv? or is this not possible
> 
> like would this work
> 
> ```
> set('os.execute', false);
> set('os.getenv', false);
> set('os.remove', false);
> set('os.rename', false);
> set('os.setlocale', false);
> set('os.tmpname', false);
> ```
That did not work sadly

Steviegt6:
> > this change would break the mod im working on since it uses functions like os.clock and os.exit would it be possible to disable only the potentially malicious functions like os.execute and os.getenv? or is this not possible
> > like would this work
> > ```
> > set('os.execute', false);
> > set('os.getenv', false);
> > set('os.remove', false);
> > set('os.rename', false);
> > set('os.setlocale', false);
> > set('os.tmpname', false);
> > ```
> 
> That did not work sadly

Would be better for a separate PR, but an alternative solution is manually reimplementing callbacks that replicate various `os` functions as needed.

Stilic:
> 

i agree

PR #4425, opened 2022-02-05
fix gf idle sometimes overriding the hairfall anim

nahiassis:


nintendofan44:
i will not close this just in case

lunacynne (approved):

PR #4811, opened 2022-02-12
add lua function isOverlapping

craftersshaft:
isColliding isn't working for me, but this does on my machine. just replaces the collide with overlap but does a good job at detecting if two objects are on top of eachother. 

These examples will only work if isOverlapping is a thing, so don't yell at me if they don't work if you haven't messed with the source code. Yell at me if the isOverlapping code itself doesn't work after being compiled.

[Simpler Example (requires you to run the game from the command line to see output)](https://github.com/ShadowMario/FNF-PsychEngine/files/8053620/script.zip)
[Complex Example (almost undertale battle system)](https://github.com/ShadowMario/FNF-PsychEngine/files/8053616/creffbattle.zip)


lemz1 (approved):
I like this since when using `isColliding('boyfriend','dad')` and moving bf to dad.x, dad is being pushed away. This doesn't happen with overlapping.
So yeah I approve
This can probably be used for some bullet or attack detection stuff.

trippinreal:
This looks interesting, could you give an example of how this would be used?

lemz1:
ezgif com-gif-maker

code:
```
function onSongStart()
    doTweenX('as', 'boyfriend', 'dad.x', 3, linear)
end

function onUpdate()
    if isOverlapping('boyfriend', 'dad') then
        setProperty('health', 0)
    end
end
```

lemz1:
Yo I just noticed ShadowMario added objectsOverlap in a commit 3 days ago

craftersshaft:
oh yeah, just saw that, Request Closed

CrazyDany:
um.. it doesn't work

PR #4913, opened 2022-02-13
makeGraphic on attachedsprite

nahiassis:
this allows you to make an **_'attached'_** sprite without it having to be an image


nintendofan44:
ay @ShadowMario

ImaginationSuperHero528:
What do you mean?

nintendofan44:
this makes you able to make a square for example without using an image file

nintendofan44:
i tried to do that without this and it gave me an error
so basically i copied and pasted this function here from the flxsprite code

trayfellow:
> this makes you able to make a square for example without using an image file

image

wdym you can do that already unless im missing something else here

frantastic24:
This isn't for Lua

HiroMizuki (approved):
useful

ShadowMario:
You can already do that though???
Screenshot_2
Screenshot_3

nintendofan44:
i think that was a bug for me
haxe is weird sometimes

PR #5572, opened 2022-02-22
Better Health Bar For People With Custom Bars

ghost:
This is for people who want to make the bars look different because they have no life essence.
Like being able to use this bar and not have the extra shit overlap it so you can't see the extra detail.
evenbetterbar
Or being able to use this bar and to have look normal instead of a box with a rounded edge.
lessbetterbar
So ye this is actually useful.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -1203,9 +1203,8 @@ class PlayState extends MusicBeatState
 	}
 
 	public function reloadHealthBarColors() {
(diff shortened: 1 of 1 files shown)


Gui-iago:
image
it doesn't quite work, sure, the custom texture is there, but the health color array for both players doesn't do much
otherwise, good pr

PrismLight:
This was broken

ProgramerGamerZone:
Why was this wont do it was so good

PR #5615, opened 2022-02-23
reimplement the non abusable os functions

lunacynne:
no difftime because you can just do `math.floor((time_2 - time_1) + 0.5)`


cyn-8:
functions currently being reimpletmented by another pr

PR #5845, opened 2022-02-26
Basic Video Sprites in Lua and Source

BendyGaming0:
Still needs to be tested but BGSprites and ModchartSprites can be attached to video bitmaps and display the bitmap onto the sprite

still needs pause, resume, loops, volume and audio support as well as not breaking the music

### DIFF
--- a/source/FlxVideo.hx
+++ b/source/FlxVideo.hx
@@ -21,7 +21,7 @@ class FlxVideo extends FlxBasic {
 	var renderOnSprite:Bool = false;
 	
 	#if desktop
-	public static var vlcBitmap:VlcBitmap;
+	public var vlcBitmap:VlcBitmap;
 	#end
 
 	public function new(name:String, ?parentSprite:FlxSprite) {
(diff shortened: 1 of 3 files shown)


BendyGaming0:
whoops, looks like there isn't a music problem, i was looking at something else and got mixed up.

GamerZoneMusic (approved):
I didn't test but I think it is very cool!

BendyGaming0:
the compiler doesn't like how im doing this so ill do it another way

BendyGaming0:
the Windows build has passed, Linux and Mac however are complaining over nothing.

TheKitBoi (commented):
The latest commit failed the compile on win, mac and linux, just a reminder for the pull requester

Jrmcwicked:
OMG WHAT I WAS LOOKING FOR THIS EVERYWHERE THANK YOU

BendyGaming0:
I'm going to close this pull, it's too buggy and I don't really have the focus for this. plus this is really only gonna be used in very specific cases where there is source code modification

PR #5903, opened 2022-02-27
Make the game safer for people with Epilepsy + for those who get motion sick & headaches easily

MeguminBOT:
Attempts to make the gameplay safer for people at risk of epilepsy/seizures or for those who get motion sick easily.

* Prevent Blammed Lights on custom songs when Flashing Lights is disabled in options.
Tested and working.

* Add Camera Shake option + Prevent Camera & Hud Shake when Camera Shake is disabled in options.
Tested and working.

LUA scripts may or may not bypass this depending on how the scripts setup.
Lets say you want to trigger ScreenShake on Blammed Lights event in LUA via `function onEvent`

    if name == 'Blammed Lights' then
      if value1 == '1' then
        cameraShake('game', '0.02', '0.1')
        cameraShake('hud', '0.02', '0.1')

In this case the screen will shake even if Camera Shake is unchecked.
However, it will prevent Blammed Lights if Flashing Lights is unchecked.

The same applies if you were to invert the code, then Blammed Lights will be triggered and Screen Shake won't.

I haven't found a good way to prevent that yet in source, if anyone has an idea, please let me know!
But in meantime, there's a way that ensures it gets blocked in the lua scripts themselves.

You can help preventing this in your own lua scripts by using `if flashingLights then` under any function that flashes colours or flashes an image on notehit etc. Same with CameraShake if this particular PR gets merged.
See one such example [HERE](https://github.com/MeguminBOT/PsychEngine-TheFatAssModCompilation/blob/main/example_mods/custom_notetypes/Static%20Note.lua)


jmxamongusmodder:
this may be cool to add to the base engine. if it is added ill port it over to pedr! (psych engine deluxe ressurection!)

lunacynne (approved):

PR #3939, opened 2022-01-25
Add instant respawn option

MisterRafers:
I think, name describes the option

PR #6454, opened 2022-03-08
Fix a small bug when retrying and quitting simultaneously

WestCraft15:
Edit: Looks like I jumped the gun a bit and didn’t notice the existing pull request for this (#3939), my bad. I like their implementation better anyway. My other patch might still be of some use, idk.

Edit 2: Might apply some other minor bug fixes later, I don't think this one alone is worth a pull request.

Original description:
I got tired of waiting for the animation to play every time I reset so I decided to add this feature. The option is in the gameplay section. This implementation shouldn't cause any problems (well, not that I can think of). Also applies a small bugfix so quitting after retrying no longer causes strange behavior.


WestCraft15:
Gonna close this pull request entirely and open a new one if/when I get around to the other bug fixes.

PR #6461, opened 2022-03-08
Installation files

DeadSyntaxdev:
Basically you open the folder than click on the files and it installs all the stuff you need for this engine besides Haxe itself.


Itz-Miles:
Now THIS is pretty cool, but is there a way to make sure these commands are run in the intended order? I dunno what'd happen if  someone doesn't happen to open the README, which hasn't been changed in this PR.

GO-Miles (approved):

XS09-dev:
> Now THIS is pretty cool, but is there a way to make sure these commands are run in the intended order? I dunno what'd happen if someone doesn't happen to open the README, which hasn't been changed in this PR.

I'll most likely number them with the name

XS09-dev:
> Now THIS is pretty cool, but is there a way to make sure these commands are run in the intended order? I dunno what'd happen if someone doesn't happen to open the README, which hasn't been changed in this PR.

Ok I numbered them. 1), 2), and 3)

GO-Miles (approved):
Seems very user friendly, especially compared to copy-pasting commands into cmd.

XS09-dev:
> Referen



> Seems very user friendly, especially compared to copy-pasting commands into cmd.

That was what I was thinking when putting it together. "it's way easier to use than copy and pasting each individual commands"

PR #6495, opened 2022-03-08
put FlxG.fixedTimestep = false back into TitleState.hx

Raltyro:


swordcube:
> what it does ?

makes the game not run like shit until you get to playstate

flxg.fixedtimestep makes the game run good no matter what

superpowers04:
From my understanding, it unlocks things like Transitions from FPS, providing the same speed animations/transitions across every device nomatter how slow or fast

PR #6512, opened 2022-03-08
Disable Editors on Story Mode

crowplexus:
kinda self explanatory


HiroMizuki (approved):
i think it kinda makes sense

PR #6637, opened 2022-03-09
Fixed OutdatedState.hx not allowing progress after restarting via Mods menu.

raf13lol:
The leftState never got reverted and it got put in a if (false). Bad move.


lunacynne (approved):

Raltyro (approved):
thank you for this, i hate being softlocked

PR #6672, opened 2022-03-10
Keeping the application alert in lua

MAJigsaw77:
This for making the user to understand what is wrong with the script


superpowers04:
The game still says what's wrong with the script, and this only appears on Mac/Linux due to a crash caused by lime

PR #6716, opened 2022-03-10
Unhide exit because it's not malicious and hide many more potentially malicious functions

superpowers04:


lunacynne (changes requested):
add these to the disable list
```haxe
os.getenv = nil;
os.remove = nil;
os.rename = nil;
package.loaded.os.getenv = nil;
package.loaded.os.remove = nil;
package.loaded.os.rename = nil;
```

lunacynne (approved):

ghost (approved):
Yes.
We can have part of os back.

ghost (changes requested):
But you should also remove some of these.
```hx
os.chdir = nil;
os.popen = nil;
os.close = nil;
os.makedir = nil;
os.makedirs =nil;
package.loaded.os.chdir = nil;
package.loaded.os.popen = nil;
package.loaded.os.close = nil;
package.loaded.os.makedir = nil;
package.loaded.os.makedirs = nil;
```
chdir changes the name of a directory/folder
popen opens a file found in a user. (if it's editable then it can also edit it, like a text editor)
close closes any open file. any. (usually used after popen but still malicious)
mkdir/makedirs makes a directory in a chosen folder

Tbh, I think you should put these in a different pr, or rename this pr.
If not, I can make my own pr.

ghost (approved):
Don't know why I have to approve it after a change.
But yea nice.
Also you missed a space with os.makedirs.

Raltyro (approved):
still kinda sad that require is blocked :sob:

PR #6922, opened 2022-03-13
Added more patten way character for playing animation

afeefflox:
basically, I want added as start frame with special animation kind of fun if I added this feature


lunacynne (changes requested):
`specialAnim` should be set after `playAnim` is called because the function sets it to `false`

afeefflox:
> `specialAnim` should be set after `playAnim` is called because the function sets it to `false`

ok I will fixed

lunacynne (approved):

superpowers04 (approved):

PR #6970, opened 2022-03-13
Add variable frame health icon support

superpowers04:
Allows health icons of any frame count to be used while still supporting vanilla icons. Anything higher than 2 frames will be treated going from Losing(left) to (Right)Winning, up to 100 frames can be defined inside of the png, no xml's required. Only caveat is that every frame has to be 150 px in width and height, if they don't comply then it'll be treated as a vanilla health icon with 2 frames

### DIFF
--- a/source/HealthIcon.hx
+++ b/source/HealthIcon.hx
@@ -30,11 +30,19 @@ class HealthIcon extends FlxSprite
 	}
 
 	public function swapOldIcon() {
-		if(isOldIcon = !isOldIcon) changeIcon('bf-old');
-		else changeIcon('bf');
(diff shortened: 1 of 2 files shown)


DM-kun (approved):
Shadow will close this, but I still approve :trollface:

crowplexus (approved):

stilic (approved):

Stilic:
btw the shit doesn't compile.

TheKitBoi (approved):

charlesisfeline (approved):

GamerZoneMusic (approved):

Theoyeah (approved):

Raltyro (approved):

PR #7089, opened 2022-03-15
Make updateHitbox for setGraphicSize and scaleObject optional

RagdollClash:
Adds a bool to setGraphicSize and scaleObject where if it's set to false, it will not call updateHitbox. Useful towards recreating Week 6's schoolEvil stage, or other stuff that scales the object but forgets to update its hitbox.

```lua
	makeAnimatedLuaSprite('bg', 'stages/schoolEvil-devilmachine/animatedEvilSchool', 400, 200);
	addAnimationByPrefix('bg', 'idle', 'background 2', 24, true);
	setScrollFactor('bg', 0.8, 0.9);
	scaleObject('bg', 6, 6, false);
	setProperty('bg.antialiasing', false);
	addLuaSprite('bg', false);
```

2022-03-15_13-37-02_PsychEngine

It is set to true by default so it doesn't break currently existing mods.

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -1087,34 +1087,34 @@ class FunkinLua {
 				}
 			}
 		});
-		Lua_helper.add_callback(lua, "setGraphicSize", function(obj:String, x:Int, y:Int = 0) {
+		Lua_helper.add_callback(lua, "setGraphicSize", function(obj:String, x:Int, y:Int = 0, updateHitbox:Bool = true) {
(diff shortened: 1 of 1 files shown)


frantastic24 (approved):
Makes sense, definitely ways better than having to use `setProperty`

ghost (approved):
You need at least 1 pull request merged. You have brilliant ideas.

stilic (approved):

PR #7988, opened 2022-04-06
Saving the last volume 

MAJigsaw77:


BeastlyGhost:
this is already on ClientPrefs.hx though

PR #7996, opened 2022-04-07
add an extra 0 to accuracy display

EyeDaleHim:
This makes it so that on the accuracy display, It adds an extra 0 when the accuracy reaches numbers that reach a length of 4 (87.50%)

So, generally, by adding a 0, this fixes the accuracy display a bit, here's a few examples:

(87.6% > 87.60%)
(44.4% > 44.40%)
(99.9% > 99.90%)

image

Not happening when it comes to those like (85.51%)
image

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -2320,7 +2320,7 @@ class PlayState extends MusicBeatState
 				{
 					if (convertedAccDisplay.charAt(i + 2) == '')
 					{
-						accuracy += '0';
+						convertedAccDisplay += '0';
 						break;
 					}
 				}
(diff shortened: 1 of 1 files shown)


guizinhomartinez:
idk not trying to be rude but this is kinda useless tho

EyeDaleHim:
> idk not trying to be rude but this is kinda useless tho

I mean, is it weird to not have that one either though?

Games like osu! and Etterna have this sort of thing.

ShadowMario:
Nah

SonicProYT:
is that really necessary to close that?

PR #8070, opened 2022-04-10
Change fps text size back to how it was before

Trayfellow:
the fps text size in the current verison seemed different so i changed the size back to how it was before 0.5.2

### DIFF
--- a/source/openfl/display/FPS.hx
+++ b/source/openfl/display/FPS.hx
@@ -46,7 +46,7 @@ class FPS extends TextField
 		currentFPS = 0;
 		selectable = false;
 		mouseEnabled = false;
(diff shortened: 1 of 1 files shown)


trayfellow:
> Ok but why?

the new fps text size kinda bothered me so i just changed it back to how it was before

Phoneguytech75:
I don't think this is a problem that affects everyone. It only affects yourself. You shouldn't have created a pull request as it doesn't change much about the game. I think you should just keep this change to your personal repo.

trayfellow:
> I don't think this is a problem that affects everyone. It only affects yourself. You shouldn't have created a pull request as it doesn't change much about the game. I think you should just keep this change to your personal repo.

fair

PR #8085, opened 2022-04-11
The hidden achievements will be hidden now :)

MAJigsaw77:


Trayfellow (approved):

Theoyeah:
This is useless, i tried and it won't work

trayfellow:
> This is useless, i tried and it won't work

wait it works for me tho

PR #8208, opened 2022-04-18
Yeah, repaired easter egg

Wither362:
Well, I activated the Easter egg, and now I can’t disable it, so I make this to restore the easter egg, and return it to the normal screen


frantastic24:
Easter eggs aren't broken, you disable them by typing the code again
This PR is useless

TheKitBoi (changes requested):
run it back

Wither362:
I mean, I typed bb, and now I can’t get gf back, but with this, it will reset the Easter egg, not disabling it

frantastic24:
You type bbpanzu not bb, close this already

frantastic24:
TitleState.hx line 469
image

PR #8341, opened 2022-04-26
Automatically grab refresh rate for first time saves

RagdollClash:
This PR makes it so any save created for the first time will have Psych automatically grab the display's refresh rate and set its framerate cap to that, instead of the user having to go into options to set it themselves.

If the monitor's refresh rate is either under or over the FPS cap range, Psych Engine will cap it.

This PR also simplifies the code for setting the game's framerate cap.

### DIFF
--- a/source/ClientPrefs.hx
+++ b/source/ClientPrefs.hx
@@ -4,6 +4,7 @@ import flixel.FlxG;
 import flixel.util.FlxSave;
 import flixel.input.keyboard.FlxKey;
 import flixel.graphics.FlxGraphic;
+import lime.app.Application;
 import Controls;
 
 class ClientPrefs {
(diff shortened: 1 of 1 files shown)


stilic (changes requested):
pls make a if else instead

ActualMandM:
wait shit yeah you can do an if else even with the compiler stuff around it, hold on

Itz-Miles:
If I have a monitor that's capped at 60Hz, wouldn't this prevent the internal framerate from being over it? I prefer the cool frames number being 120Hz, despite not always testing source on a 120Hz capable display.

ActualMandM:
It'll only get the monitor's refresh rate when it first creates the save.

lunacynne (approved):

ShadowMario:
Revert the change in GraphicsSettingsSubState.hx because it's unnecessary and make it save the framerate to ClientPrefs.framerate

ActualMandM:
Reverted the GraphicsSettingsSubState change, though I'm confused about the saving to `ClientPrefs.framerate` part. Wouldn't it already be doing that if I'm editing the `framerate` variable within ClientPrefs?

ActualMandM:
Closing due to no response...

Raltyro:
sob

PR #8417, opened 2022-04-30
made a simple thing that may prevent event crashes

nahiassis:


nintendofan44:
nvm it doesnt prevent crashes
but its cool and can say to the user if an event doesn't exist

Stilic:
> nvm it doesnt prevent crashes but its cool and can say to the user if an event doesn't exist

you should open a dialogue box instead of tracing shit in the console

nintendofan44:
i will do that later or use luatrace

BombasticTom:
LOL

PR #8449, opened 2022-05-02
bpm fix

voxovistired:
firstly, why are we getting the value from an array and not a float 💀


Stilic:
no

nintendofan44:
this is hella useless
its only a trace
its used for console debugging
and the way it was before is the way it should be

voxovistired:
shits useless you right lmao

PR #8454, opened 2022-05-03
Fix combo count in EditorPlayState

RagdollClash:
Quick and simple bugfix, this fixes the combo count in EditorPlayState being off by one and only showing past 10, making it function the same as PlayState. This also fixes the combo count going over 9999 in both PlayState and EditorPlayState.


BombasticTom:
I don't think they left that as an accident

BombasticTom (approved):

PR #8464, opened 2022-05-03
Add HTML alerts for web builds.

ghost:
For every FlxG warn in PlayState.hx there is a js function that makes an alert if you are compiling to HTML5. Seems pretty self-explanatory doesn't it?

# Tell me if there are any other FlxG warns you want me to put a js function into.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -38,6 +38,9 @@ import flixel.util.FlxSort;
 import flixel.util.FlxStringUtil;
 import flixel.util.FlxTimer;
 import haxe.Json;
+#if js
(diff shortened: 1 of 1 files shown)


BombasticTom:
maybe make it so that it adds warning messages if you're using debug mode for haxe

PrismLight:
> maybe make it so that it adds warning messages if you're using debug mode for haxe

That's already used with `FlxG.log.warn()`

BombasticTom (approved):

PrismLight:
Surprised this hasn't been closed yet.

PR #8606, opened 2022-05-12
.

Joalor64:
Maybe there could be an extra watermark in the Main Menu??
 I just think it could be a good idea.


Stilic:
trash

Joalor64GH:
well i tried

PR #3572, opened 2022-01-15
BPMs in chart editor are no longer capped at 339

Jaduplex:
removed completely arbitrary limitation. now you can make songs up to 1000 bpm which is way more than youd ever need so enjoy

PR #8625, opened 2022-05-13
No more BPM Capping.

Joalor64:
This is kind of a duplicate of #3572, but better. Now the max BPM in the chart editor is 10000. 
(This is my first time making pull requests, okay?)

### DIFF
--- a/source/editors/ChartingState.hx
+++ b/source/editors/ChartingState.hx
@@ -470,7 +470,7 @@ class ChartingState extends MusicBeatState
 		clear_notes.color = FlxColor.RED;
 		clear_notes.label.color = FlxColor.WHITE;
 
-		var stepperBPM:FlxUINumericStepper = new FlxUINumericStepper(10, 70, 1, 1, 1, 339, 1);
+		var stepperBPM:FlxUINumericStepper = new FlxUINumericStepper(10, 70, 1, 1, 1, 10000, 1);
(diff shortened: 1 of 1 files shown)


PrismLight:
jsfdjnfsdjk;adfphoiusdgr7deywatquwrekkkkkkkkkkkkkkdhfn4wj;oiph35fyurjmks,l;
The hardest song.

Joalor64GH:
imagine somebody made that an actual mod lol that would be funny i think

PrismLight:
> imagine somebody made that an actual mod lol that would be funny i think

Ik. But why 10k nobody even needs that.

Joalor64GH:
i know it's a bit too much

BombasticTom:
> > imagine somebody made that an actual mod lol that would be funny i think
> 
> Ik. But why 10k nobody even needs that.

you forgot about dave and bambi porters, right?

lunacynne (approved):

Wither362:
Wait, I tested this at it crashed, this isn’t good for low pcs, instead of 10000 use 1000 bpm

Wither362 (changes requested):
I approve this, but I have tested this, and for low pcs isn’t good, better if the max bpm is 1000 instead of 10000

Joalor64GH:
you're right

PR #8658, opened 2022-05-15
New Menu BG.

Joalor64:
I simply just used the background for the charter from @gedehari's Izzy Engine and then used a recolor tool online for the different menu bgs.
(Still my first time making pull requests.)
(Also, ignore the other commit. That was for another pull request.)
(Also, sorry if menuDesat.png is a little too dark.)


SrtHero278:
this might not get merged since Shadow wants to keep this Engine as vanilla as possible.

Joalor64GH:
oh wait your right

XperiorTR:
you can add this through the mods folder because assets from the mod folder override vanilla files

Joalor64GH:
ohh okay

Lokitot:
you should add this an option, like dark mode

Joalor64GH:
why did i not think of that?

Lokitot:
> why did i not think of that?

idk homie

PR #8664, opened 2022-05-15
mult speed is now functional

te-agma-at:
had to make it that way, no other choice than to do so :\


lunacynne (approved):

ShadowMario:
It already works though?

PR #8665, opened 2022-05-15
Scroll speed for individual notes

te-agma-at:
hopefully this will work

### DIFF
--- a/source/Note.hx
+++ b/source/Note.hx
@@ -79,6 +79,7 @@ class Note extends FlxSprite
 	public var noAnimation:Bool = false;
 	public var hitCausesMiss:Bool = false;
 	public var distance:Float = 2000; //plan on doing scroll directions soon -bb
+	public var multSpeed:Float = 1; //you can now change individual note scroll speed using lua :O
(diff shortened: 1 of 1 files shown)


te-agma-at:
> > hopefully this will work
> 
> it wont? you didnt update funkinlua and edit code to even make it work. theres more then just adding a variable kid.

where though?

xChillax:
You need to make the notes use that variable to make it work, otherwise as it stands, that variable does nothing.

Stilic:
also shadowmario don't want that

xChillax:
Oh, and why is that? I'm surprised how many things are being denied despite giving more creativity to mods.

(besides the point that this right here is not even properly done)

Stilic:
> Oh, and why is that? I'm surprised how many things are being denied despite giving more creativity to mods.
> 
> (besides the point that this right here is not even properly done)

idk

hayfii:
Do you even know what coding is ☠️

ShadowMario:
This doesn't do shit

PR #8726, opened 2022-05-21
Update MainMenuState.hx

NimuFurrySoftUwU:


BeastlyGhost:
why would you even do that in the first place? this is the base engine

BoloVEVO:
When you don't know how to clone a repository and mod directly the base engine instead. *Fatherless face*

BeastlyGhost:
> When you don't know how to clone a repository and mod directly the base engine instead. _Fatherless face_

:skull:

SrtHero278:
This is a dumb thing.
There's no reason to have a mods menu if there's no mods.

PR #8907, opened 2022-06-03
Update hxCodec Libraries

crowplexus:
this allows for videos on linux and a handy skip button for them (linux is disabled for the time being)


EyeDaleHim (approved):
I like this a lot

jigsaw-4277821:
If I remember correctly the height is bugged on hxCodec so this is a bad addition

stilic (commented):
Why these files are still here?

BeastlyGhost:
> Why these files are still here?

because the game won't compile without them, that's why shadowmario copied them to the source folder in the first place

Stilic:
They are already included in the Haxelib package. We just need to install it.

Now, I recommand you to remove these files and to add a note inside the README.

BeastlyGhost:
> They are already included in the Haxelib package. We just need to install it.
> 
> Now, I recommand you to remove these files and to add a note inside the README (and also to install HxCodec using Haxelib if it isn't done already).

won't do, I want to have the files there to make double sure this will work, I did test compiling without them and I just don't want to have new issues regardling compiling issues because the vlc folder doesn't exist, if it works, then it works, leave it as is

Stilic:
> > They are already included in the Haxelib package. We just need to install it.
> > Now, I recommand you to remove these files and to add a note inside the README (and also to install HxCodec using Haxelib if it isn't done already).
> 
> won't do, I want to have the files there to make double sure this will work, I did test compiling without them and I just don't want to have new issues regardling compiling because the vlc folder doesn't exist, if it works, then it works, leave it as is

Sorry but I tried back before my PR for add hxCodec was closed with the Haxelib and it perfectly worked.
So, seriously, remove these files please.

BeastlyGhost:
> You all know then the height is glitched in hxCodec right?

I know, it's not my job to fix it though, and even then if poly ever fixes it, no PR needs to be made

Stilic:
> You all know then the height is glitched in hxCodec right?

fix it then lol

jigsaw-4277821:
> 

I tryed almost a month and nothing

jigsaw-4277821:
@BeastlyGhost  did

BeastlyGhost:
made another one just to wrap it up on one commit

PR #8940, opened 2022-06-05
Adding a achievement notification instead of the old thing

MAJigsaw77:
is tested don't think about i didn't did that, just take a look yourself @ShadowMario
Screenshot (153)
it looks similar to indie cross achievement notification system


An-undertale-fan:
So this is the indie cross achievement thingie
Look great the normal psych achievement thing is boring 😒

Swapper1:
This look better than the old one that can wait for 2 second until the next song or send back to Story/Freeplay menu. Funkin' Approved.

ghost (approved):

source/MainMenuState.hx
```diff
@@ -251,7 +230,7 @@ class MainMenuState extends MusicBeatState
 					});
 				}
 			}
-			#if desktop
+			#if MODS_ALLOWED
```
RafPlayz69YT, line 233: Uh, it seems kinda dumb to do this, what if you want to edit characters quickly?????

jigsaw-4277821: You are right

PR #8946, opened 2022-06-06
Allow "Add Camera Zoom/Zooming" to work before the opponent sings

ghost:
I really need it, if anyone don't need it just do setProperty('camZooming', false); or idk???


hayfii:
bruh

PR #8948, opened 2022-06-06
Remove FlxFlicker from FlashingState

kinoko87:
It's pretty contradictory to flash lights kinda fast if a user doesn't want them


hayfii:
how about if (ClientPrefs.flashing) ?

ShadowMario:
White letters won't kill someone, they aren't a flashing color, it's just letters on a black screen.

PR #8991, opened 2022-06-08
Added isStagePixel and curStage for if statements

rodney528:


ImaginationSuperHero528:
For Lua... cause why not?

rodney528 (commented):
Now to wait for it to be accpected or declined.

source/FunkinLua.hx
```diff
@@ -189,6 +189,8 @@ class FunkinLua {
 		set('noResetButton', ClientPrefs.noReset);
 		set('lowQuality', ClientPrefs.lowQuality);
 		set("scriptName", scriptName);
+		set('isPixelStage', PlayState.isPixelStage); // BTW idrk if these will actually work properly since I don't fully know what I'm doing with haxe lmao I only have some understandings like how I had to add the "PlayState." part
+		set('isPixelStage', PlayState.curStage);
```
frantastic24, line 193: What

frantastic24:
I don't think this is necessary tbh

CrusherMods:
can't you just use
```lua
if getProperty('isPixelStage') then
```
???

frantastic24:
You can, which is why I don't think this is necessary or even useful since both variables are very rarely used

Mastercopper2346:
You forgot to change isPixelStage to curStage

ImaginationSuperHero528:
> You forgot to change isPixelStage to curStage

Wdym I meant for both to be there?

CrusherMods:
> > You forgot to change isPixelStage to curStage
> 
> Wdym I meant for both to be there?

image

source/FunkinLua.hx
Raltyro: what??

ImaginationSuperHero528:
Fixed an error opss!

DabbinDude316B (approved):

An-undertale-fan:
If curStage is kinda useless since you can write the code in the stage lua :v

ImaginationSuperHero528:
Changed to a custom branch.

ImaginationSuperHero528:
Not dead go to #9350

PR #9091, opened 2022-06-17
Blah blah VideoHandler not MP4Handler blah blah

stilic:
bruh the thing changed just a month ago if i remember.
@BeastlyGhost why??


BeastlyGhost:
psych is using the old version, it's named MP4Handler there

ShadowMario:
:skull::moyai:

PR #9115, opened 2022-06-19
fix mod directory on endSong

Raltyro:
why is this in the endsong anyway


Raltyro:
oopsie i make some mistake

PR #9121, opened 2022-06-19
Update CharacterEditorState.hx

CasualCypress:


Stilic:
what this does?

jmxamongusmodder:
i think it adds offset fixing and does other things?

afeefflox:
:/ guys I think  he copy code from blantados (that was so DUMB I never seen)

PR #9168, opened 2022-06-22
Lua - onSpawnStrum function

BombasticTom:
This function activates when a strum is added (basically at the beginning of a song)
It's really useful despite the fact that some players make cutscenes in the beginning of a level

It has 3 parameters: `position`, `player` and `id`

position - index number of the strum in `strumLineNotes`
player - to which player does the strum belong to? (0 is the opponent while 1 is the player)
id - public strum's noteData (basically what strum is it, left down up or right)

Script example - Placing the notes just like in vanilla fnf:

```
function onSpawnStrum(position, player, id)
        if not middlescroll then
	        setPropertyFromGroup('strumLineNotes', position, 'x', getPropertyFromClass('Note', 'swagWidth') * id + 50 + (screenWidth / 2) * player)
        end
end

```
(haven't tested this script so I am not certain if this would work)


BombasticTom:
if i broke something by making the player variable global in strumnote.hx, i'll make another merge

stilic (approved):

PR #9216, opened 2022-06-25
Clearing the ram in freeplay 

MAJigsaw77:
It really needs to clear the memory bruh


ShadowMario:
That was purposely removed.

PR #9327, opened 2022-06-28
fix score text memory leak

crowplexus:
that's about it

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -2868,9 +2868,10 @@ class PlayState extends MusicBeatState
 		setOnLuas('curDecStep', curDecStep);
 		setOnLuas('curDecBeat', curDecBeat);
 
-		scoreTxt.text = 'Score: ' + songScore + ' | Misses: ' + songMisses + ' | Rating: ' + ratingName;
-		if(ratingName != '?')
-			scoreTxt.text += ' (' + Highscore.floorDecimal(ratingPercent * 100, 2) + '%)' + ' - ' + ratingFC;
+		scoreTxt.text = 'Score: ' + songScore
+		+ ' | Misses:' + songMisses
+		+ ' | Rating: ' + ratingName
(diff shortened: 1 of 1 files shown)


Raltyro (approved):
Isn't this just the same like the previous pr?

Paracosm-Daemon:
what's the memory leak here?

Paracosm-Daemon:
> > what's the memory leak here?
> 
> text was constatly being generated

i don't really understand what you mean by that, like adding onto the text in the if statement causes a memory leak even though the text is being reset before it?

BeastlyGhost:
> > > > > what's the memory leak here?
> > > > 
> > > > 
> > > > text was constatly being generated
> > > 
> > > 
> > > i don't really understand what you mean by that, like adding onto the text in the if statement causes a memory leak even though the text is being reset before it?
> > 
> > 
> > putting "+=" on _update(elapsed)_ makes the text constatly generate, so if the rating name was not "?", it would generate the Rating Percent text every frame, this fixes it by making the text always be there by default but hidden if the rating name is "?", rather than being generated afterwards
> 
> though, wouldn't that operand be the exact same as the logic gate? you're still **add**ing a string onto an existing string, so wouldn't that be equivalent to +=?, just shorter?

luckily not, since the text is hidden behind the variable and not being added as you press a note, it's like putting an object as "visible = false" and making it visible afterwards, since the object was not visible before, it was not affecting the game memory, by doing that the memory shouldn't increase drastically if that same object became visible by, say, `if (curStep == blah)`

BeastlyGhost:
plus adding objects every frame is not a good thing either, I will probably make a pull request in the future so score bar updates won't depend on `super.update(elapsed)`, rather, the text won't update unless `goodNoteHit`, `noteMiss`, or `noteMissPress` was triggered, basically making update useless for updating such text

Starveldt (approved):

BeastlyGhost:
closing because I'd rather make it the **good** way

PR #9373, opened 2022-06-29
It shouldn't be in a string 

MAJigsaw77:


Raltyro (approved):
so true

ShadowMario:
Yes it should

MAJigsaw77:
.....

MAJigsaw77:
You broke lua again as always

PR #9433, opened 2022-07-01
ifCurAnim

rodney528:
It's `getProperty('sprite.animation.curAnim.name')` and `getProperty('sprite.animation.curAnim.finished')` basically.

Example use...
```lua
if ifCurAnim('boyfriend', 'pre-attack', true) then
    playAnim('boyfriend', 'attack')
end
```
ifCurAnim(obj:String, anim:String, done:Bool = false)

`obj` is either a luaSprite or character.
`anim` your desired animation.
`done` if true then it's after the animation finishes playing. If false then just when the animation plays in general.

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -1533,19 +1533,17 @@ class FunkinLua {
 		});
 		
 		Lua_helper.add_callback(lua, "getCurAnim", function(obj:String, anim:String, done:Bool = false) {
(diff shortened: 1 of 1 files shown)


ImaginationSuperHero528:
Forgot to test!

frantastic24:
Seems a little confusing, why is it called getCurAnim if it doesn't return curAnim

PR #9440, opened 2022-07-02
Create Psych Installation Stuff.bat

rodney528:
Basically everything ready for first time use! Idk if curtain things should be there btw.


ImaginationSuperHero528:
Maintainers can edit btw if there are errors.

PR #9112, opened 2022-06-18
Update hxCodec (Video Sprites)

crowplexus:
this update allows for BG Video Sprites, looping videos, hardware accelerated videos and fixed callbacks

PR #9448, opened 2022-07-02
Fixed Mp4 video handler error

JonnycatMeow:


Raltyro (commented):
Someone (BeatlyGhost) already did this, it's currently in draft because haxelib for hxCodec isn't updated yet
https://github.com/ShadowMario/FNF-PsychEngine/pull/9112

trayfellow:
this is cause the current version of hxcodec psych uses is the haxelib version, if you want to compile psych engine make sure ur using the haxelib version instead of the git version of hxcodec, we should wait hxcodec to update first then this can be changed

BeastlyGhost:
this error doesn't need to be fixed, as Psych Engine uses the stable branch of hxCodec

image
the stable branch uses MP4Handler, which is basically the old version of VideoHandler, simply install from stable rather than from the git branch with `haxelib install hxCodec`

EyeDaleHim (changes requested):
Question, are you on 0.6.2?

ItsyourboyJonnycat:
i think so

PR #9464, opened 2022-07-02
Faster playback rates

Starmapo:
Adds the unused gameplay modifier that changes the song's playback rate, up to 2.5x speed. It uses `FlxG.timeScale` so basically everything will automatically be sped up, including timers, tweens, and animations, without having to do it manually.
Slower playback rates are _possible_ but they cause audio errors near the end of the song so I opted to remove them for now.

Would've included a video but there's an upload size limit.

### DIFF
--- a/source/Conductor.hx
+++ b/source/Conductor.hx
@@ -23,6 +23,7 @@ class Conductor
 	public static var songPosition:Float=0;
 	public static var lastSongPos:Float;
 	public static var offset:Float = 0;
+	public static var playbackRate:Float = 1;
 
 	//public static var safeFrames:Int = 10;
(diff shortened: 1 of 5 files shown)


crowplexus (approved):
i present to you https://8mb.video

Raltyro (approved):
CHIPMUNKS

dummmmdum (approved):

BalancedLight (approved):
now this is epic

KadeDev:
> so basically everything will automatically be sped up, including timers, tweens, and animations, without having to do it manually.

idk man. I think a better version of this would be using OpenAL like shadow is already doing and then just increasing/decreasing the bpm and make everything based on the bpm.

Speeding up everything sounds like a terrible idea that'll cause a lot of shit to break/look jank.

PR #9525, opened 2022-07-05
Pausing every timer/tween not just modchart_timers/tween

lemz1:
Basically this pr add the feature which pauses every timer/tween instead of just the modchart ones.
This is useful for people that use source.


TheKitBoi:
the fastest pr close on the west

PR #9527, opened 2022-07-05
trace file that is returning null errors and disable fonts folder embedded on the executable + add setup file

crowplexus:
TL;DR - useful for debugging and embedding the fonts path is really not needed anymore, as psych has custom font support on mods

---------------------------------------

this PR enables file outputs on files that are "returning null NOOOO"
output should be something like
```
oh no NOTE_assets.xml is returning null NOOOO
```
(NOTE_assets is an example)

the fonts folder was embedded by default on the base game, Psych however, has support for custom fonts, this might be handy in case you wanna change, say, the FPS Counter with source code using a custom font for example, since the fonts folder will be generated once the game is compiled, then you could just do something like ``Paths.font("pixel.otf")`` for example


adamsaustria:
it is still doing this error

BeastlyGhost:
> it is still doing this error

this is not to fix the error, because it's not an error in the first place, it is simply to print the **file** that is returning the error

Raltyro (approved):

rodney528 (approved):

JuniorNovoa1 (approved):
fucking please

dummmmdum (approved):

MemeHoovy (approved):

rodney528 (approved):

JorosTheModderPH (approved):

source/Paths.hx
```diff
@@ -355,7 +355,7 @@ class Paths
 			localTrackedAssets.push(path);
 			return currentTrackedAssets.get(path);
 		}
-		trace('oh no its returning null NOOOO');
+		trace('oh no ' + key + ' is returning null NOOOO');
```
Itz-Miles, line 358: numbers can't explain how ingenious this single addition is, but words can.

approved/10

GO-Miles (approved):

BeastlyGhost:
will reopen later maybe idk

PR #9675, opened 2022-07-13
Add a 'black screen function'

DabbinDude316B:
why not

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -3908,6 +3908,19 @@ class PlayState extends MusicBeatState
 		}
 	}
 	#end
+	
+	public var screenGoesBlack:Bool = true;
+	function blackScreen(?charsGoAway:Bool = true, ?screenGoesBlack:Bool = true) { // this can also be used for winter horrorland
+		if (screenGoesBlack) {
+			camGame.visible = false; //Setting both cams go invisible makes it go black? idk why lol
+			camHUD.visible = false;
+			camOther.visible = false; //why not?
+		} else {
+			camGame.visible = true; //Setting them back
+			camHUD.visible = true;
+			camOther.visible = true;
+		}
+	}
 
 	public function KillNotes() {
 		while(notes.length > 0) {
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -1263,7 +1263,7 @@ class PlayState extends MusicBeatState
 					if(gf != null) gf.playAnim('scared', true);
 					boyfriend.playAnim('scared', true);
 
-				case "winter-horrorland":
+				/*case "winter-horrorland":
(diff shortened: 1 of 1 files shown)


Mastercopper2346:
> why is `screenGoesBlack` both a variable and a parameter?

Wdym

source/PlayState.hx
```diff
@@ -3908,6 +3908,20 @@ class PlayState extends MusicBeatState
 		}
 	}
 	#end
+	
+	public var screenGoesBlack:Bool = true;
+	function blackScreen(?screenGoesBlack:Bool = true) { // this can also be used for winter horrorland
```
steviegt6, line 3913: There is no reason to declare both a public field and function parameter with the same name here.

source/PlayState.hx
```diff
@@ -1263,7 +1263,7 @@ class PlayState extends MusicBeatState
 					if(gf != null) gf.playAnim('scared', true);
 					boyfriend.playAnim('scared', true);
 
-				case "winter-horrorland":
+				/*case "winter-horrorland":
```
steviegt6, line 1266: Winter Horrorland won't have the desired effect since the code is commented out without a proper replacement.

source/PlayState.hx
```diff
@@ -3908,6 +3908,20 @@ class PlayState extends MusicBeatState
 		}
 	}
 	#end
+	
+	public var screenGoesBlack:Bool = true;
+	function blackScreen(?screenGoesBlack:Bool = true) { // this can also be used for winter horrorland
+		if (screenGoesBlack) {
+			camGame.visible = false; //Setting both cams go invisible makes it go black? idk why lol
+			camHUD.visible = false;
+			camOther.visible = false; //why not?
```
steviegt6, line 3917: Why not what? If you feel the need to say something like this, I believe it would be better to explain yourself.

source/PlayState.hx
```diff
@@ -3908,6 +3908,20 @@ class PlayState extends MusicBeatState
...
+		} else {
+			camGame.visible = true; //Setting them back
+			camHUD.visible = true;
+			camOther.visible = true;
+		}
+		setOnLuas('blackScreen', screenGoesBlack)
```
steviegt6, line 3923: A bit of a nitpick, but you could also just make the function parameter `blackScreen` as well for consistency.

steviegt6 (changes requested):

source/PlayState.hx
Mastercopper2346: I mean why not make this

source/PlayState.hx
Mastercopper2346: I know, I wanted to be careful so the engine wouldn’t get confused

source/PlayState.hx
Mastercopper2346: I’ll do it it soon

PR #9839, opened 2022-07-21
Change health via difficulty

DabbinDude316B:


Itz-Miles:
You might want to merge this functionality with the gameplay changers, resulting in gameplay settings per difficulty.

rodney528 (approved):
Meh

source/Note.hx
```diff
@@ -265,6 +265,18 @@ class Note extends FlxSprite
...
+			switch(CoolUtil.difficultyString()) {
+				case 'Easy':
+					hitHealth = 0.06;
+					missHealth = 0.02;
+				case 'Normal':
+					hitHealth = 0.02;
+					missHealth = 0.04;
+				case 'Hard':
+					hitHealth = 0.01;
+					missHealth = 0.03;
```
Intancote, line 278: Don't know if this is intentional or not but is the missHealth supposed to be lower on Hard than Normal?

Mastercopper2346: Yep, it’s intentional. Press 3 notes to make up for that one miss.

Intancote: Alr then

source/Note.hx
```diff
@@ -265,6 +265,18 @@ class Note extends FlxSprite
 				prevNote.updateHitbox();
 				// prevNote.setGraphicSize();
 			}
+			
+			switch(CoolUtil.difficultyString()) {
```
ImaginationSuperHero528, line 269: Uh... what about custom difficulties?

Mastercopper2346: Wait

Itz-Miles:
You should still put this in options or gameplay changers. This PR changes gameplay not just for the engine, but every mod you'll ever play on psych, making some harder mods impossible to beat (also mentioning that this will just break most mod compatibility in general)

source/Note.hx
Mastercopper2346: k i fixed it :))))

Mastercopper2346:
fine.. i'll work harder on it.

frantastic24:
This is pretty useless considering you can just use gameplay modifiers and/or setProperty

PR #9847, opened 2022-07-21
Change minimum delay to -500 miliseconds

### LINKED ISSUE #8465: Allow setting a negative note offset
### What feature do you want to get added on the **base** engine?

Currently (up to and including the latest version) it is not possible to set a note offset to a negative value. This makes it harder to play on setups with a lower delay than usual.

For example, I experience about -20 ms of delay with the game, which is not a problem I can fix in Psych Engine. There are other engines which do not have this limitation (including Kade Engine, I believe) where the delay can be set properly.

### To test your sight, and reliability, please select the option of what should NOT be requested.

Stage Editor, 6K+ support, and winning icons.

junetried:
-500 is probably unrealistically low, but it makes the bar nice and even. More to the point, this fixes #8465.

I expected this to involve more pain, like changing an unsigned integer type or changing how the configuration is saved. It's none of that.

### DIFF
--- a/source/options/NoteOffsetState.hx
+++ b/source/options/NoteOffsetState.hx
@@ -31,7 +31,7 @@ class NoteOffsetState extends MusicBeatState
 	var dumbTexts:FlxTypedGroup<FlxText>;
 
 	var barPercent:Float = 0;
-	var delayMin:Int = 0;
+	var delayMin:Int = -500;
 	var delayMax:Int = 500;
 	var timeBarBG:FlxSprite;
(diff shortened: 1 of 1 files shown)


Intancote (approved):

MemeHoovy (approved):

lunacynne (approved):

junetried:
I assume there isn't any good reason why this was closed either, is there? Fine. I like compiling every version of Psych Engine I want to use anyway.

PR #9991, opened 2022-07-31
Simple Project.xml changes.

Joalor64:
It's just simple. Not much.
It makes it so that the executable filename is different depending on the platform.
Also, tjson.

### DIFF
--- a/Project.xml
+++ b/Project.xml
@@ -2,8 +2,12 @@
 <project>
 	<!-- _________________________ Application Settings _________________________ -->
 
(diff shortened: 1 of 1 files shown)


DillyzThe1 (commented):
While I don't exactly understand but seem against the name change for the game per platform (but also not for), why exactly is "tjson" randomly imported?

Joalor64GH:
Because in main.yml, there is this line:
```
haxelib install tjson
```

So, why not import it in Project.xml?

Playertrip777:
I'm not sure why tjson is necessary, it doesn't seem to affect us.

ShadowMilky:
> Because in main.yml, there is this line:
> 
> ```
> haxelib install tjson
> ```
> 
> So, why not import it in Project.xml?

i swear i dont think the source code uses tjson at all

Joalor64GH:
oh

PR #10113, opened 2022-08-08
Cleanup comments in GraphicsSettingsSubState.hx

doinkers:
Remove parameter explanations since they are already explained above in the Anti-Aliasing option


ShadowMario:
useless

PR #10134, opened 2022-08-10
Extra animations block sing anims

DabbinDude316B:
a


Raltyro (commented):
gotta be handy, but can you turn on actions in your fork so we can see if its compiling?

EyeDaleHim:
This is an... okayish PR, but although it;s appreciated but I feel like this goes somewhere as a suggestion instead.

PR #10150, opened 2022-08-11
Probably better of clearing unused memory (In my opinion)

Just-Feeshy:
The cached assets will stay until no longer in the PlayState. I think this is useful because there would be an asset that has been cached that’ll be created after the create function has been run, or a mod could be using the same asset the song after.


EyeDaleHim:
Good method, but I've seen better

Just-Feeshy:
hehe

PR #10248, opened 2022-08-19
Slow down pixel strum hits from 24fps to 12fps to match other pixel note animations

RedstoneRuler:
Very simple change

### DIFF
--- a/source/StrumNote.hx
+++ b/source/StrumNote.hx
@@ -65,11 +65,11 @@ class StrumNote extends FlxSprite
 				case 0:
 					animation.add('static', [0]);
 					animation.add('pressed', [4, 8], 12, false);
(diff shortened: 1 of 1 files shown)


MemeHoovy (approved):

TheKitBoi (approved):

Parsellol (approved):

RedstoneRuler:
Yeah, I couldn't figure out a straightforward way to word it. The pixel presses are all at 12fps due to how few frames there are for them, but the pixel hits are 24fps, except for the up arrow, which is 12.

RedstoneRuler:
The inconsistency makes me think it was probably a mistake. (It's like this in base FNF too btw)

PR #10281, opened 2022-08-22
the better ways easier

afeefflox:
bruh you don't need two Animation Indices loop and non loop just made into one bruh

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -1789,14 +1789,10 @@ class FunkinLua {
 			}
 		});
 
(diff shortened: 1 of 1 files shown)


ManMaster707:
While I understand where you're coming from, I think you should keep the AddAnimationByIndicesLoop function for backwards compatibility with older mods.

Raltyro:
tis wont work cuz linc luajit doesnt supports 6 or more arguments, they just will be nil

rodney528 (changes requested):
> this wont work cuz linc luajit doesnt supports 6 or more arguments, these will just become null

What raltyro said.

ShadowMario:
bro didn't test his own commit...

PR #10283, opened 2022-08-23
Tiny oversight that probably doesn't effing matter

rodney528:
I think that should be a 3... not a 4.

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -171,7 +171,7 @@ class FunkinLua {
 		set('botPlay', PlayState.instance.cpuControlled);
 		set('practice', PlayState.instance.practiceMode);
 
-		for (i in 0...4) {
+		for (i in 0...3) {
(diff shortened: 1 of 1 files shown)


ImaginationSuperHero528:
Fun Fact I did this all off of my phone :D surprisingly didn't run into much difficulty. Since I only really changed one number.

Raltyro:
i think it does matter, haxe loop is kinda odd
when you do 0...4 it will end at 3, and 0...3 will end at 2

lemz1:
4 is right.
for loops in haxe work differently than other programming languages.
in this case ***i*** will equal to
0,1,2,3
so yeah close this pr

ImaginationSuperHero528:
> 4 is right. for loops in haxe work differently than other programming languages. in this case _**i**_ will equal to 0,1,2,3 so yeah close this pr

Oh... ok.

PR #10319, opened 2022-08-26
Optimize some of the Note Prefix Code

crowplexus:
instead of setting them up like this

```Haxe
animation.addByPrefix('blah', 'blah');
```

we instead use arrays, making the code optimized in terms of how many lines it takes


Mastercopper2346:
Got merged instantly

PR #10461, opened 2022-09-07
move base chart parsing to a separate class

crowplexus:
NOTES TO TAKE: while it would be good to have them there, events are still present on playstate, but in case you wanna edit base chart data, go to the ChartParser class, as I moved it to there

from my testing this didn't break anything, initially events were also going to be on the chart parser, however this breaks them as a whole, so until I (or someone else) discover how to move them without breaking them, they are going to stay on PlayState

### DIFF
--- a/source/ChartParser.hx
+++ b/source/ChartParser.hx
@@ -1,12 +1,129 @@
 package;
 
+import flixel.math.FlxMath;
+import flixel.FlxG;
+import Section.SwagSection;
+import sys.FileSystem;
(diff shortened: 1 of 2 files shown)


MemeHoovy (approved):

DabbinDude316B (approved):

iym1337 (approved):

Parsellol (approved):

BeastlyGhost:
losing because it's useless ngl

PR #10621, opened 2022-09-22
Texture null stuff for splashes

rodney528:
Basically, it checked only for "null" before and I added "or ''"


ImaginationSuperHero528:
uhhhh

PR #10712, opened 2022-09-30
WIP: Refactoring file handling and mod

marius851000:
I want to add better handling for mods (due to... some ideological preference for clean mods instead of source port, with some exception... But that’s not important. Also, this started due to wanting to support mod in ~/.local/share/something on Linux (and whatever the ApplicationPath is on Windows)).

It’ll probably take a few days until I have something good with that, but here is the overview of what I plan:

Adding two class, a class ModInfo, representing a mod (and the related pack.json) and a modsList, representing the list of avalaible mod, loading or not.

ModInfo will be a pure (with no static) class, but modsList will have multiple, mostly:

a static list of Path (String) indicating folders where there are one subfolder per mods (sorted by priority list)
a static list of ModInfo, indicating the list of loaded mods
It’ll also have a function to load a list of ModInfo of all known mods, whether they are enabled or not.
(I still have a few decision to take regarding API, but I’ll take them).

I made that WIP, as I will be interested in comments regarding the way it’s made, both about the implementation and the design.
Also, it’s my first contribution here (and the first code I write in haxe, althought I’m already experienced with a bunch of other languages), so if I did anything wrong, I’ll be interested to know that too.

(also, recreated https://github.com/ShadowMario/FNF-PsychEngine/pull/10711 as I forgot to mark it as Draft)


MemeHoovy:
I'm pretty sure this breaks backwards compatibility

marius851000:
I'll try not to break backward compatibility. The only problem there should be with backward compatibility is about modsList.txt

The kinda solution I had is first to load the one in the user folder, then read the one on the game folder. When saving, save it to the same modsList that goes with the folder the mod is in (if in the user folder, then the user file, if the game mods folder, then on the binary directory).

The issue I still need to solve with this is mod ordering. I have some idea about weight, bit I'll have to thing a bit more about it.

marius851000:
*otherwise, already have an old breaking patch that simply change the mods folder to be a user folder*

marius851000:
So... I have an hard time finding a perfect solution (and even if I find one, it will end up being a complex CRDT. May be appropriate in some context, but it is likely too much here). Here are the main element that I consider:

- It needs to support multiple installation. I guess may happen that certains mods are distributed with the Psych Engine, with the mods itself being like a normal mod in that PsychEngine distribution (ideally, with a pack.json)
- It also need to support installation of mods independantly of where the software is installed, like what most other software does (save in the user config folder or whatever)

So, I have imagined two solution:

1. If there is a modsList.txt in the binary folder only use this one, and only make mods of this specific Psych Engine installation available (that is, those that are in the mods folder in the game). Otherwise, load or create one in the user’s config folder, and use the mods that are in the user’s mod folder.
2. If there is a modsList.txt in the binary folder, load it, but also loads the mods in the user folder, with a lower priority. Re-ordering and enabling/disabling is then saved in the modsList.txt in the binary folder, not changing the one in the user’s config folder. Otherwise, change the user’s config modsList.txt, and make both the mods in the mods game folder and the user’s mod folder avalaible.

After having wrote all that, the first solution seems to be the best in regard to simplicity of implementation, simplicity of use by the users and the modder (someone sharing a custom PsychEngine can just add a modsList.txt to the mod, and it won’t make use of the user’s mod) and being the less likely to have bug. I’ll start implementing this way.

marius851000:
It also have the benefit of not changing how the users use it if they already have mods installed, and just update the installation. New installation will however need users to install their mods in their mods folder, and would thus need an update to mod installation instruction.

marius851000:
So... I figured I had some misunderstanding about how some things work. That is, a week does not need to be added in modsList.txt to work (but it can be excluded). I’ll find a solution taking this into account.

marius851000:
I will make mods load weeks by default, but not be added in the GlobalModList.

marius851000:
*with the appropriate change to the mods selection menu*

marius851000:
> I will make mods load weeks by default, but not be added in the GlobalModList.

That is, modsList.txt will be in the this format now (conceived so not to break compatibility with older version):

<modName>|enabled|globalEnabled (default to enabled if already in the file, false otherwise)

marius851000:
I’m hesitant in making these separate. On one hand I’m pretty sure that, if the user add a mod, they want to play their song, on another side, I’m not that sure he want it be able to change the main menu background or the main menu music...

For now, I’ll just default to enable it globally when it is enabled, and disable is globally if it is disabled, so every enabled mod is also globally enabled. It should just a matter of changing some assignement and change the UI in the mod list menu to change it.

marius851000:
I’ll make it so entry in modsList.txt that reference a mod that doesn’t have a folder are removed from it.

marius851000:
Also, to whoever wrote this:
doing
```
while (i < array.length) {
    if (array[i].something) {
        array.remove(array[i]);
        continue;
    }
}```

won’t work as expected, as it work something like this

you have `array = [ 1, 2, 3, 4 ]`
you execute a first iteration, the first element (`array[0]`) is kept and `i=1`

on second iteration, the second element (`array[1]`) is removed, so now, `array = [ 1, 3, 4]` and `i=2`

on the third iteration, the third element is kept (`array[2]`) is kept. But the third element is actually 4 now. So 3 is skipped.

(and on the fourth iteration, it does out of bound read).


(I saw it here. It seems this is done by @ShadowMario (I hope you don’t me doing this)). I’ll fix this with this patch.

marius851000:
So... There is something I want to change, but that will cause backward incompatibility in certain case (in a way that shouldn’t be problematic):

While the mods folder itself contain subfolders for mods, it is itself a mod. I plan to remove this, so mods only contains subfolders for other mods.
This will probably only cause issues for those sharing mods AND the engine in a future version that would include this change. To update, they would just need to create a folder for their mod, and put it in the mod folder (then, it’ll be auto-activated as global mod, and everything will work correctly).

I also plan of checking for the presence of a `disable_mod_menu` file, that, if present, will disable the Mods entry in the main menu.

marius851000:
After more thought, I instead choose to use `load_all_mods.txt`, which load all mod in the mods folder, except the example mod. (and thus, will ignore the `modsList.txt` file).

marius851000:
> After more thought, I instead choose to use `load_all_mods.txt`, which load all mod in the mods folder, except the example mod. (and thus, will ignore the `modsList.txt` file).

That’s actually a bad idea to ignore the modsList.txt, as it risk to mess load ordering. I’ll go back on the idea of `disabled_mod_menu`. (mods are enabled globally by default anyway)

marius851000:
Here is the change for the credit. I added a `CreditEntry` class, that store information about a credit entry. Also, two other bug fixe:
1. The background color is while instead of black if there is no color defined
2. The icon is not displayed if there are no such one
(not sure if these were present before, or introduced by my changes...)

marius851000:
Another change (again. There really is a scope creep here. But it’s fun refactoring code).

I’ll make the preload path to be considered as a (global) mod for the purpose of iterating over mods (except it won’t be saved in modsList, won’t appear in the mod selection menu and will always be last)

marius851000:
So... Regarding mod priority... I’ll probably standarize on:

current mod, then all the globals mods, then the game assets (which is technically the last global mod)

marius851000:
I’ll also add more abstraction between the embedded assets and the file system files. Note that I’ll make sure to maintain compatibility with JS. (thought the current version may not be, as I don’t test for this currenctly)

marius851000:
One potential benefit of that would be to allow mod in the web just by embedding them and a modsList.txt

marius851000:
The last PR allow to list sub directories of embedded ressources. As such, we don’t need the weeklists.txt, stagelists.txt and characterlists.txt. And it get me nearer to running mods on html5 (even if that’s not the original goal. Working on Psych Engine is really fun, with a good codebase, but still not perfect, so I can improve it).

marius851000:
Closing it due to lack of motivation to continue it. Maybe I’ll continue it later, who know. But fell free to fork it if you want to continue it.

marius851000:
Okay. I fell like continuing it now, so I’ll re-open and fix what remains to fix (and rebase it on main)

Issue #10737, opened 2022-10-02
Memory leak of the lua stack on main


marius851000:
### Describe your bug here.

Hello. I was doing some development related to mods handling, but I hit this nasty bug. I suspect it to be a memory leak, given it atually previously work (but not sure at which revision. Will do a bisection.

This happen at revision ff917e6b1de454d27dc7d95c7169e41ff9154ee8 on the Dusk Till Down mod. It seems to only happen if I let the intro video fully play, and crash near the end of the song.

It crash when there is about 22 second remaining on the song.

### Command Prompt/Terminal logs (if existing)

```bash
source/Discord.hx:18: Discord Client starting...
source/Discord.hx:68: Discord Client initialized
source/Discord.hx:25: Discord Client started.
source/LoadingState.hx:162: Setting asset folder to shared
source/Conductor.hx:135: new BPM map BUDDY [{ stepTime => 144, stepCrochet => 78.9473684210526, bpm => 190, songTime => 18000 }]
source/FunkinLua.hx:2772: initializing haxe interp for: mods/Dusk Till Dawn/stages/castle.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/stages/castle.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/custom_notetypes/GF Sing Alt.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/custom_notetypes/Pinkie Sing.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/custom_events/Laser.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/data/dusk-till-dawn/cutscene.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/data/dusk-till-dawn/gameCamera.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/data/dusk-till-dawn/note camera.lua
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/data/dusk-till-dawn/precache.lua
source/Paths.hx:362: oh no its returning null NOOOO
source/FunkinLua.hx:108: lua file loaded succesfully:mods/Dusk Till Dawn/data/dusk-till-dawn/script.lua
play cutscene
[...]

PR #10740, opened 2022-10-02
lua: fix stack memory leak

marius851000:
`Lua.pushnil(lua);` doesn’t help, as it seems the function result use a different stack than the one stored in the class.

In addition, it had the bad side-effect of adding element that were never removed. In the extreme case (that I met), it would add new entry every frame, until lua make the game crash (see: https://github.com/ShadowMario/FNF-PsychEngine/issues/10737#issuecomment-1264648248 )

I also make a PR against the forked linc_luajit used here, that make so returning null corretly transfer into nil to Lua: https://github.com/nebulazorua/linc_luajit/pull/8

I’m a bit worried that would mean that people who have an up-to-date linc_luajit will have a different result than those who don’t, but I don’t see how to make it better.

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -309,13 +309,11 @@ class FunkinLua {
 			var shader:FlxRuntimeShader = getShader(obj);
 			if (shader == null)
 			{
-				Lua.pushnil(lua);
 				return null;
 			}
 			return shader.getBool(prop);
 			#else
 			luaTrace("getShaderBool: Platform unsupported for Runtime Shaders!", false, false, FlxColor.RED);
-			Lua.pushnil(lua);
 			return null;
 			#end
 		});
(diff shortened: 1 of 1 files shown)


MemeHoovy (approved):

TheKitBoi (approved):

stilic (approved):

Raltyro (approved):
So how people with different linc_luajits will do if this "pr" (not the one in nebulazorua/linc_luajit currently) will do if this got pushed?

marius851000:
> So how people with different linc_luajits will do if this "pr" (not the one in nebulazorua/linc_luajit currently) will do if this got pushed?

As in, what change will there be before and after this PR is merged (the one PsychEngine)?

(I have no idea why a fork is used. How! a conflict. Let’s fix that)

DM-kun (approved):

source/FunkinLua.hx
```diff
@@ -490,13 +478,13 @@ class FunkinLua {
...
-				return;
+				return null;
 			}
 			if(funcName==null){
 				#if (linc_luajit >= "0.0.6")
 				LuaL.error(lua, "bad argument #2 to 'callScript' (string expected, got nil)");
 				#end
-				return;
+				return null;
 			}
```
Raltyro, line 488: R

aaaproooo (approved):

lunacynne (approved):

marius851000:
Rebased on master.

PR #10841, opened 2022-10-13
gfNote Variable

rodney528:
Basically, grab if a note press is from a gfNote.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -4531,12 +4531,15 @@ class PlayState extends MusicBeatState
 		RecalculateRating(true);
 
 		var char:Character = boyfriend;
+		setOnLuas('gfNote', false);
 		if(daNote.gfNote) {
+			setOnLuas('gfNote', true);
 			char = gf;
 		}
-
+		
 		if(char != null && !daNote.noMissAnimation && char.hasMissAnimations)
 		{
+			setOnLuas('gfNote', false);
 			var animToPlay:String = singAnimations[Std.int(Math.abs(daNote.noteData))] + 'miss' + daNote.animSuffix;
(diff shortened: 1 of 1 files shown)


MemeHoovy (approved):

TheKitBoi (approved):

frantastic24:
What is the point of this
You can just use the note type

BeastlyGhost:
this accomplishes like nothing at all, that's why the Notetype variable is set on luas anyways

ImaginationSuperHero528:
> What is the point of this
> You can just use the note type

Bro you can set multiple notes with this var so no.

ImaginationSuperHero528:
> this accomplishes like nothing at all, that's why the Notetype variable is set on luas anyways

Could elaborate on the "that's why the Notetype variable is set on luas anyways" part?

BeastlyGhost:
> Could elaborate on the "that's why the Notetype variable is set on luas anyways" part?

gfNotes are notetypes too, one could simply use that in order to locally make a `gfNote` variable on a lua script, or just use the noteType variable itself

Stilic:
sorry but it's literally useless

ShadowMilky:
this is useless cuz you can just use the notetype

ImaginationSuperHero528:
> this is useless cuz you can just use the notetype

This is to grab the var that is the key component of the "GF Sing" note.

frantastic24:
there already is a var, it's called note type
again, this is useless

ImaginationSuperHero528:
So... this is actual garbage.

ImaginationSuperHero528:
So... this is actual garbage.

rodney528:
Ah, the good ol days. When I was a **fucking idiot**.

PR #10881, opened 2022-10-17
Playable Character check for characters other than BF

NerdiestAlpaca:
Adds an additional check for check_player in the character editor that automatically sets the variable to true if the selected character file name has the suffix "-player" as opposed to only doing that when the file name starts with BF


NerdiestAlpaca:
(idk if this code is like actually good or not but it at least functioned when I put it in a mod that I'm working on so there's that

RandomGnome34 (approved):

MemeHoovy (approved):

CoolingTool (approved):
i think instead of adding the if statements, `daAnim.startsWith('bf')` should just be replaced with `daAnim.startsWith('bf') || daAnim.endsWith('-player')` but thats just a stylistic change, other than that the pr is good

PR #10903, opened 2022-10-19
Make compiling with haxe 4.1.5 impossible

MemeHoovy:
I took that from Hope engine, they're underrated as hell.
https://github.com/skuqre/Hope-Engine

And also made it so you couldn't compile if your not using the git version of Linc_luajit (Use nebs fork for the love of god people, can't you fucking read instructions?)
https://github.com/ShadowMario/FNF-PsychEngine/blob/fe5104c1cf927e990e43a82009854f8258cc282b/README.md#L9

### DIFF
--- a/Project.xml
+++ b/Project.xml
@@ -191,7 +191,7 @@
 	<!-- I was trying to make it so it'd give you an error if your not using NebulaZoruas fork of linc_luajit but I couldn't figure it out. -->
 	<error value="Seems like you're not using NebulaZoruas github fork of luajit, please use it instead." if="${linc_luajit == 0.0.4}" />
 
-	<error value="Seems like you're using Haxe ${haxe}, please update!" if="${haxe == 4.1.5}" />
+	<error value="Seems like you're using Haxe ${haxe}, please update!" if="${haxe == 4.1.5}" unless="linux"/>
 
 	<!-- _________________________________ Custom _______________________________ -->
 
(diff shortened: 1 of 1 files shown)


APRO716 (approved):

Itz-Miles (approved):

NotNimu (approved):

MemeHoovy:
> I don't like this feature at all; A Chrome OS (which is a type of Linux if you are on one with the support) user is forced to use Haxe 4.1.5.

can't you use haxe 4.2.5 on chrome os?

MemeHoovy:
let me do something rq

coolfren:
> Sudo Apt Install haxe only installs 4.1.5

lol get the [haxe ppa](https://launchpad.net/~haxe/+archive/ubuntu/releases)

TheKitBoi (approved):

XS09-dev (approved):

XS09-dev:
I only approve it because well we need it. But I feel like it's worthless against a expert programmer, I think you should make it a separate .hx file and then add it onto the .gitignore just so it's harder to access and get rid of if even possible to access.

MemeHoovy:
> I only approve it because well we need it. But I feel like it's worthless against a expert programmer, I think you should make it a separate .hx file and then add it onto the .gitignore just so it's harder to access and get rid of if even possible to access.

nah

XS09-dev:
> > I only approve it because well we need it. But I feel like it's worthless against a expert programmer, I think you should make it a separate .hx file and then add it onto the .gitignore just so it's harder to access and get rid of if even possible to access.
> 
> nah, works fine the way it is

Eh. What do I know? Pretty much 0 about .xml files. So yea probs is.

PR #11000, opened 2022-10-25
Various Stuff from Lunar Engine

KirbyKid256:
These would be triggered SPECIFICALLY when you exit the current week rather than just when the script is destroyed (which happens whenever you end, restart, or exit the song).

### DIFF
--- a/source/ClientPrefs.hx
+++ b/source/ClientPrefs.hx
@@ -134,7 +134,7 @@ class ClientPrefs {
 		FlxG.save.flush();
 
 		var save:FlxSave = new FlxSave();
-		save.bind('controls_v2' #if (flixel < "5.0.0"), 'ninjamuffin99' #end); //Placing this in a separate save so that it can be manually deleted without removing your Score and stuff
+		save.bind('controls_v2', CoolUtil.getSavePath()); //Placing this in a separate save so that it can be manually deleted without removing your Score and stuff
 		save.data.customControls = keyBinds;
 		save.flush();
 		FlxG.log.add("Settings saved!");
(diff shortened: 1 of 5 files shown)


MemeHoovy (commented):
iirc I think lua has a destroy thing, where a script is triggered whenever the script is destroyed, so isn't that everytime a song ends/exits? in that case this makes this useless

KirbyKid256:
> iirc I think lua has a destroy thing, where a script is triggered whenever the script is destroyed, so isn't that everytime a song ends/exits? in that case this makes this useless

Yes, but "everytime" includes restarting the song, ending the song, exiting the song, going back in time with Skip Time in Charting State, entering the Chart Editor, and entering the Character Editor. This addition would make it so that it's ONLY triggered when you're basically transitioning from PlayState to StoryMenuState or FreeplayState. And while the onExitSong is triggered the moment you start to exit the song, you can also use exitingsong with onDestroy when the script is destroyed because you're exiting the song.
For example: `function onDestroy()
if getProperty("exitingsong") then
--[[code here]]
end
end`

CoolingTool (changes requested):
should opening the chart editor really count as exiting? also i think `exitingsong` should be camel cased like pretty much every other variable

KirbyKid256:
> should opening the chart editor really count as exiting? also i think `exitingsong` should be camel cased like pretty much every other variable

For the first one, I could make a different callOnLuas like "onOpenChartEditor" or something like that. For the second one, I somehow read the endingSong variable as all lowercase which was why I made exitingSong all lowercase. I'll definitely fix that one.

TheKitBoi (approved):

BombasticTomDev (approved):
Decent changes, could help modders in the future.

CoolingTool (approved):

PR #11080, opened 2022-11-01
GameJolt Support

GamerPablito:
Since Indie Cross existed, everyone wants to make mods to be compatible with GameJolt.
So I thought: "Why don't give it a shot and implement a custom client for it by myself?"

I know everyone wants this, so I think it would be a good idea to implement it into PE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,10 @@ APIStuff.hx
 art/build_x32-officialrelease.bat
 art/build_x64-officialrelease.bat
 art/test_x64-debug-officialrelease.bat
+
 ### GameJolt Private Stuff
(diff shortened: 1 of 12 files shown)


DabbinDude316B:
Ye, I’ll add this to my mod for credit

DabbinDude316B:
I’ve been wondering how to do it since _aflac_.

GamerPablito:
@DabbinDude316B Remember, this is a draft, is not finished yet, but I'm glad you liked the idea

MemeHoovy:
ShadowMario wants this engine to be kept as close to the vanilla engine as possible though

GamerPablito:
Oh, ok, there's no problem at all :)

PR #11100, opened 2022-11-02
Freeplay

Manuel1260:
I have detected that when putting freeplay and when there are many loaded mods it takes time to open the freeplay menu and it is somewhat annoying because it takes about 6 or 5 minutes to load something that in the previous ones like 6.1 does not take long to load I would like them to solve it


MemeHoovy:
okay, but you open an issue, not a stupid ass PR

PR #11108, opened 2022-11-03
Fix for hold note behaviour when strum direction is changed

UncertainProd:
When the strum line direction is changed using lua like so:
```lua
setPropertyFromGroup('strumLineNotes.members', 0, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 1, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 2, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 3, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 4, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 5, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 6, 'direction', 100)
setPropertyFromGroup('strumLineNotes.members', 7, 'direction', 100)
```
It causes the hold notes to break apart as they remain vertical even when the direction is changed.
This pr should resolve that

Before:
brokenholds

After:
fixedholds

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -3205,6 +3205,10 @@ class PlayState extends MusicBeatState
 					}
 
 					var angleDir = strumDirection * Math.PI / 180;
+					
+					if(daNote.isSustainNote)
+						daNote.angle = strumDirection - 90;
+					
 					if (daNote.copyAngle)
 						daNote.angle = strumDirection - 90 + strumAngle;
 


MemeHoovy (approved):
tails are a bit fucked but whatever, It's not really noticable

TheKitBoi (approved):

Luisinhi010 (approved):

lemz1 (approved):

source/PlayState.hx
```diff
@@ -3205,6 +3205,10 @@ class PlayState extends MusicBeatState
 					}
 
 					var angleDir = strumDirection * Math.PI / 180;
+					
+					if(daNote.isSustainNote)
+						daNote.angle = strumDirection - 90;
+
```
ImaginationSuperHero528, line 3211: Try using note.distance for trails may help.

rodney528 (approved):

PR #11118, opened 2022-11-04
Fix 32 Bit Exporting Error

crowplexus:
fuck you lime.


MemeHoovy (approved):
idk what the hell happened for 32 bit builds to not work

BeastlyGhost:
> idk what the hell happened for 32 bit builds to not work

stuff changed on lime so now you have to append a different value while compiling

PR #11180, opened 2022-11-08
reasons why my dementia hasn’t affected me as a person

penissucker77:


CaptainBaldi:
image
fire

penissucker77:
> image
> fire

Ok now I’m actually curious what would that even affect

PR #11198, opened 2022-11-10
Gameplay changer support by mod

endergreen12:
I am using a translation site, so I apologize if my English is not correct.

## Description

With this support, it is possible to add gameplay changers by adding files to custom_gamechangers folder in the mods folder.

You can also retrieve its value from lua. (getGameplayChangerValue)

The filename of the .txt file added to custom_gamechangers becomes the title on the settings screen, and the contents of the file become the value type and the tags inside the game and the property of the value.

Custom game changer selections are saved in the save data.

## Examples to add

For example, to add a gameplay changer that reduces health when dad hits a note, add a file named "dad note hits reduce health.txt" with contents like "bool, dadReduceHP, false" and add code to reduce health by making a conditional decision using
"if getGameplayChangerValue( 'dadReduceHP' ) then" in the lua file.

HP reduction settings can also be added by float.

image

----------------------------------------------------------------------------------------------------------------------------------------

This is what it looks like when you actually add it:

https://user-images.githubusercontent.com/90076182/201365697-1396556b-f8bc-420d-9cb5-51b082a05fef.mp4

### DIFF
--- a/source/GameplayChangersSubstate.hx
+++ b/source/GameplayChangersSubstate.hx
@@ -140,6 +140,10 @@ class GameplayChangersSubstate extends MusicBeatSubstate
 								case 'string':
 									var option:GameplayOption = new GameplayOption(fileToCheck, tag, type, contentArray[0], contentArray);
 									optionsArray.push(option);
+
+								case 'bool':
+									var option:GameplayOption = new GameplayOption(fileToCheck, tag, type, contentArray[0]);
+									optionsArray.push(option);
 							}
 						}
 					}
--- a/source/GameplayChangersSubstate.hx
+++ b/source/GameplayChangersSubstate.hx
@@ -117,10 +117,12 @@ class GameplayChangersSubstate extends MusicBeatSubstate
 					var path = haxe.io.Path.join([directory, file]);
 					if (!FileSystem.isDirectory(path) && file != 'readme.txt' && file.endsWith('.txt')) {
 						var fileToCheck:String = file.substr(0, file.length - 4);
-						var tag:String = File.getContent(path);
+						var contentArray:Array<String> = File.getContent(path).split(",");
+						var type:String = contentArray[0];
+						var tag:String = contentArray[1];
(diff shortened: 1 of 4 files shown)


DetectiveBaldi (approved):
nice

TheKitBoi (approved):

Parsellol (approved):

charlesisfeline (approved):

aaaproooo (approved):

source/GameplayChangersSubstate.hx
```diff
@@ -94,6 +98,66 @@ class GameplayChangersSubstate extends MusicBeatSubstate
...
+							contentArray.remove(type);
+							contentArray.remove(tag);
+							switch(type)
+							{
+								case 'int' | 'float' | 'percent':
+									var option:GameplayOption = new GameplayOption(fileToCheck, tag, type, contentArray[0]);
+									option.scrollSpeed = Std.parseFloat(contentArray[1]);
+									option.minValue = contentArray[2];
+									option.maxValue = contentArray[3];
+									option.changeValue = contentArray[4];
```
UncertainProd, line 139: Just to be safe, I think it's better to wrap these lines with Std.parseFloat() like below, instead of passing the string directly
```
option.minValue = Std.parseFloat(contentArray[2]);
option.maxValue = Std.parseFloat(contentArray[3]);
option.changeValue = Std.parseFloat(contentArray[4]);
```
There was a bug that I was facing when using Int values in the options and this seemed to fix it

UncertainProd (changes requested):
Really cool feature overall, I just have a few minor suggestions

source/GameplayChangersSubstate.hx
Endergreen12: thanks for your suggestions, i have incorporated all your advice into the code

UncertainProd (approved):

BombasticTomDev (approved):

CoolingTool (approved):

rodney528 (approved):

PR #11252, opened 2022-11-17
line optimize

HiroMizuki:
there's 2 lines declaring the same thing since it could be in only 1 line


Raltyro (approved):

mateus-x02 (approved):
good pr

MateuzinhoX02:
some things in the engine may use switches and cases, and other things use ram for no reason, but, it's all very optimized, so congratulations to those involved

PR #11294, opened 2022-11-22
Fix compiling error

Heropowerbrine:
Zoom is not used anymore, if it is not removed the game cannot compile


TheKitBoi (approved):

BeastlyGhost:
this causes a compiling error for however's using flixel 4, you could just have done [this](https://github.com/BeastlyGhost/Forever-Engine-Feather/blob/d4d3ea08e0c11f90bd35862c58f2c447d23b9f55/source/Main.hx#L176)

Heropowerbrine:
Closing as this won't work with older flixel versions, which was fixed in #11299

PR #11314, opened 2022-11-25
callFromObject & callFromClass

Blossomical:
cause it would be so useful for some cases
and also added stampSprite since it could also be useful

### DIFF
--- a/source/FunkinLua.hx
+++ b/source/FunkinLua.hx
@@ -2017,6 +2017,14 @@ class FunkinLua {
 				PlayState.instance.modchartSprites.remove(tag);
 			}
 		});
+		
+		Lua_helper.add_callback(lua, "stampSprite", function(sprite:String, brush:String, x:Float, y:Float) {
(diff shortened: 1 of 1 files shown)


charlesisfeline (approved):

system32unknown:
how `callFromClass` is works?

system32unknown (changes requested):
change `arguments:Array<Dynamic>` to `?arguments:Array<Dynamic>`,
then put `
    if (arguments != null) 
        arguments = [];
`

Raltyro (approved):

TheKitBoi (approved):

DisIsNOt (approved):

system32unknown (approved):

rodney528 (approved):

rodney528 (approved):

MaybeMaru (approved):

YanniZ06 (changes requested):
Why would you name the variables like that pleaseeee
how hard is it to make proper readable variable names

PR #11388, opened 2022-12-03
Migrate the Base Engine to Flixel 5

crowplexus:
a few days ago, [Flixel 5.0.0](https://github.com/HaxeFlixel/flixel/releases/tag/5.0.0) was released, and with it, many MANY things changed
some of those changes include Default Save File location, the way the `FlxGame` creation works, along with other changes,

* it may be helpful to [read this](https://github.com/HaxeFlixel/flixel/wiki/Flixel-5.0.0-Migration-guide) for more information

this Pull request migrates the base engine to the newest version of Flixel, along with keeping support for older versions, enables the fonts folder inside `assets`, enables `hxcpp-debug-server` for debug builds, adjust variables on `Main` and changes how saves are binded

this will hopefully help with [current known issues](https://github.com/ShadowMario/FNF-PsychEngine/issues/11369)

PR #11409, opened 2022-12-04
Fix error due to Flixel 5.0 removing zoom from FlxGame

superpowers04:


superpowers04:
https://github.com/ShadowMario/FNF-PsychEngine/pull/11388 ﻿

PR #11471, opened 2022-12-11
fix -null on using chart editor on boot up

raf13lol:
this makes it so if it gets a null difficulty (which it usually does on boot-up and not loading any songs), it will just not have a file suffix

### DIFF
--- a/source/CoolUtil.hx
+++ b/source/CoolUtil.hx
@@ -17,28 +17,26 @@ using StringTools;
 
 class CoolUtil
 {
-	public static var defaultDifficulties:Array<String> = [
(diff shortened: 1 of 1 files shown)


SillyVyxia (approved):

TheKitBoi (approved):

Mokona59 (approved):

brujfnf:
hi raf 😁

RafPlayz69YT:
hi

ThePlank:
hi raf 😁

RafPlayz69YT:
hi

PR #11580, opened 2022-12-26
Add proper HTML5 Support

SanicBTW:
This should kind of improve load times when preloading as well as adding proper support for HTML5 targets.

### DIFF
--- a/Project.xml
+++ b/Project.xml
@@ -79,7 +79,7 @@
 	</section>
 	
 	<section if="NO_PRELOAD_ALL">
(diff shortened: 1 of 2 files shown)


Sezei:
More curious about why you'd clear the cache only if it's HTML5

SanicBTW:
I honestly don't understand the way caching works on HTML5 with OpenFL but I remember getting some asynchronous error when trying to play songs back in 0.5.2h, when I tested the engine without any changes it worked fine sooo I don't really know

PR #11608, opened 2022-12-29
subclass for Character.hx

Joalor64:
Boyfriend.hx is now a Subclass or something.


ShadowMilky:
could be useful when theres errors like this
image
but thats about it lol

Joalor64GH:
nvm it was already deleted in `experimental`

PR #11625, opened 2022-12-31
fix chart editor playback sustain bug

TehPuertoRicanSpartan:
fixes that one bug that triggers you by vortex editor mode ending a sustain note animation too early or late when playback speed is changed (minor fix btw)


RedstoneRuler (approved):
Nice, this is something that always bothered me.

UmbratheUmbreon (approved):

PR #11636, opened 2023-01-01
bpmTxt updating when needed

Burgerballs:


tposejank:
image
man this still gonna update every frame + no affect to performance

Burgerballs:
> image man this still gonna update every frame + no affect to performance

That is because the bpmtxt needs to update every frame when the song is playing, or else it wont show the curbeats and stuff

tposejank:
> > image man this still gonna update every frame + no affect to performance
> 
> That is because the bpmtxt needs to update every frame when the song is playing, or else it wont show the curbeats and stuff

then wtf is this for

Burgerballs:
So it doesn't update when the information stays the same. Such as when the song is paused or if you are using the UI tabs.

PR #10861, opened 2022-10-15
Bring back softcoded achievements v2

ghost:
This PR brings back the scrapped softcoded achievements.
To make softcoded achievements, you'll need a .json and .lua file.
Here's an example.

So this is an example .json file: 
```json
{
	"name": "Road to Ten Million",
	"desc": "Gain 10 Million scores in total.",
	"save_tag": "ten_million",
	"hidden": false,
	"clearAchievements": true,
        "week_nomiss": null,
        "lua_code": null,
	"global": null,
	"index": -1
}
```
And there is the .lua file (optional, you can just put the code to **lua_code**).
```lua
function onUpdate(elapsed) 
  if getProperty("songScore") >= 10000000 then
    giveAchievement("ten_million");
  end
end
```

### JSON
Mandatory variables:

- **name** is obviously the name of the achievement.
- **desc** is the description of your achievement.
- **save_tag** is the save tag of the achievement, also the file name of the achievement icon.
- **hidden** is the visibility of your achievement. Remains invisible until it's unlocked if true.
- If **index** is null or -1, gets added to end of the achievements.

Optional variables:

### **WARNING**: Using these optional variables is dangerous and using these can cause unstability. Use carefully.
**clearAchievements** clears all of the achievements if true. **SHOULD BE USED ONLY ONCE IN A MODPACK.**
**global** is the most dangerous one. If you want to set **global**, remember all of the other variables will get ignored.
If **global** is not null, it replaces the WHOLE achievements. So a .json file with **global** should be like this:
```json
{
	"name": "",
	"desc": "",
	"save_tag": "",
	"hidden": false,
	"clearAchievements": false,
	"index": null,
        "global":  [
            ["Road to Ten Million", "Gain 10 Million scores in total.", "ten_million", false],
            ["Road to a Billion", "Gain 1 Billion scores in total.", "one_billion", false]
        ]
}
```
This way, achievements gets replaced with **global**. First element is name, second is description.
Third is the save tag and icon file name, fourth is the visibility until unlocked.
### JUST LIKE **clearAchievements**, **global** SHOULD BE ONLY USED ONCE IN A MODPACK.

**week_nomiss** is the week name, just put the file name of the week to it to make a "No miss" achievement.
**week_nomiss should always end with "nomiss"!  For example, if your week name is "week8" and you want
to make a beating week with no misses achievement, just write "week8_nomiss" to week_nomiss!**

**lua_code** is the code of the achievement. If you do not want to create a lua file, just put your code in it!

### LUA
[...]

PR #11651, opened 2023-01-04
Softcoded achievements

ghost:
This PR is a rewrite of a shitty PR I made before, #10861.
If you are using the old PR for your mods, please switch to this one.
Also, this is going to be a draft for a while because it may have bugs I don't know about. 

Changes:

- Json files with **global** in it must have a lua file.
- Added **song**, more information below.
- Achievement lua script will ignore of your return attempts, please don't try to return anything with them.
- ***giveAchievement*** works with any function now (https://github.com/ShadowMario/FNF-PsychEngine/pull/10861#issuecomment-1356893529), and only achievement scripts can run it.
- Softcoded achievements will work only on current mod folder, and preload folder.
- ***giveAchievement*** will return useful logs, use ***debugPrint*** to see the logs.
- No miss week achievements are supported for real now.

### UPDATED DESCRIPTION:
To make softcoded achievements, you'll need a .json and .lua file.
Here's an example.

So this is an example .json file: 
```json
{
	"name": "Road to Ten Million",
	"desc": "Gain 10 Million scores in total.",
	"save_tag": "ten_million",
	"hidden": false,
	"clearAchievements": true,
        "week_nomiss": null,
        "lua_code": null,
	"global": null,
	"index": -1,
        "song": ""
}
```
And there is the .lua file (optional, you can just put the code to **lua_code**).
```lua
function onUpdate(elapsed) 
  if getProperty("songScore") >= 10000000 then
    giveAchievement("ten_million")
   -- will return "Unlocked Achievement ten_million" if this achievement is not unlocked
   -- if it is already unlocked, it will return "Achievement ten_million is already unlocked!"
   -- if the achievement does not exist, it will return "Achievement ten_million does not exist"
  end
end
```

### JSON
Mandatory variables:
(ALL OF THE VARIABLES HERE ARE IGNORED IF ***global*** IS NOT NULL!)
- **name** is obviously the name of the achievement.
- **desc** is the description of your achievement.
- **save_tag** is the save tag of the achievement, also the file name of the achievement icon.
- **hidden** is the visibility of your achievement. Remains invisible until it's unlocked if true.
- If **index** is null or -1, gets added to end of the achievements.
- If **song** is not null, your achievement will be only be unlockable in that specific song.

Optional variables:

### **WARNING**: Using these optional variables is dangerous and using these can cause unstability. Use carefully.
**clearAchievements** clears all of the achievements if true. **SHOULD BE USED ONLY ONCE IN A MODPACK.**
**global** is the most dangerous one. If you want to set **global**, remember all of the other variables will get ignored.
If **global** is not null, it replaces the WHOLE achievements. So a .json file with **global** should be like this:
```json
{
	"name": "",
	"desc": "",
	"save_tag": "",
	"hidden": false,
	"clearAchievements": false,
	"index": null,
        "song": null,
        "global":  [
            ["Road to Ten Million", "Gain 10 Million scores in total.", "ten_million", false],
            ["Road to a Billion", "Gain 1 Billion scores in total.", "one_billion", false]
        ]
}
```
This way, achievements gets replaced with **global**. First element is name, second is description.
Third is the save tag and icon file name, fourth is the invisibility until unlocked.
### JUST LIKE **clearAchievements**, **global** SHOULD BE ONLY USED ONCE IN A MODPACK.

**week_nomiss** is the week name, just put the file name of the week to it to make a "No miss" achievement.
**week_nomiss should always end with "nomiss"!  For example, if your week name is "week8" and you want
to make a beating week with no misses achievement, just write "week8_nomiss" to week_nomiss!**

**lua_code** is the code of the achievement. If you do not want to create a lua file, just put your code in it! **lua_code** is ignored if ***global*** is set.

### LUA
Lua is nothing special, though there are differences.
Just use **giveAchievement(*Achievement's save tag*)** to unlock the achievement.
If achievement is already unlocked or achievements doesn't exist or the lua script is not an achievement script,
**giveAchievement** is ignored.

### Example Images
Using **clearAchievements**:
image

Without using **clearAchievements**:
image

With **index** being 0:
image

With using **global**:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,24 +34,9 @@ jobs:
           sudo apt-get install libvlc-dev
           sudo apt-get install libvlccore-dev
           haxelib setup ~/haxelib
-          haxelib install hxcpp > /dev/null
-          haxelib install lime
-          haxelib install openfl
-          haxelib --never install flixel
-          haxelib run lime setup flixel
-          haxelib run lime setup
-          haxelib install flixel-tools
-          haxelib install flixel-ui
-          haxelib install flixel-addons
-          haxelib install tjson
-          haxelib install hxjsonast
-          haxelib install hxCodec
-          haxelib git linc_luajit https://github.com/nebulazorua/linc_luajit
-          haxelib install hscript
-          haxelib git hscript-ex https://github.com/ianharrigan/hscript-ex
-          haxelib git discord_rpc https://github.com/Aidan63/linc_discord-rpc
-          haxelib install hxcpp-debug-server
-          haxelib list
+          haxelib install hxcpp > /dev/null --quiet
+          haxelib install hmm --quiet
+          haxelib run hmm install
       - name: Create Version Tag
         run: echo "${{github.run_id}}" > VERSION
       - name: Compile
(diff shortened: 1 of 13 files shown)


TheKitBoi (approved):

charlesisfeline (approved):

DM-kun (approved):

Joalor64 (approved):

aaaproooo (approved):

Luisinhi010 (approved):

crowplexus (approved):
this should probably be renamed to "Softcoded Achivements" as a general term, works perfectly out of the box

DisIsNOt (approved):

Joalor64GH:
I should ask. How does `lua_code` work?

TheWorldMachinima:
> I should ask. How does `lua_code` work?

What do you mean?

Joalor64GH:
Like, are we supposed to add the name of the script or do something else??

TheWorldMachinima:
> Like, are we supposed to add the name of the script or do something else??

Put the code to *lua-code*, if you don't want to create a file for it.

PR #11713, opened 2023-01-14
fix strum animations for different playbacks

Unholywanderer:
this just fixes the weird animation reset times the strums can have for botplay/opponent when the playback rate is different, bit like how it was in the chart editor 

hopefully it doesn't break anything


UmbratheUmbreon:
if you were to do this, why not just put the playbackRate division in strumPlayAnim?
like:
image

PR #11752, opened 2023-01-20
use splice parameter in `remove();` and remove splashes when they are killed

Leather128:
probably helps with memory usage in a very slight amount somewhere and seems to not break the game at all so i think it's a reasonable thing to accept


voiddevv:
very cool 😎

CaptainBaldi:
leather128 w

PR #11811, opened 2023-01-29
Filepath tracing, custom notes in the chart editor, and XML crash prevention

GO-Miles:
# **Overview**
**_This request changes Note.hx_** so the engine attempts to reload custom notes in-game and in the chart editor with a valid texture rather than reloading with a blank string, which can be overridden by the noteskin and cause crashes.

**_It also changes Paths.hx_** to trace the expected filepath of an asset returning null.

## The bugs:

- Calling `reloadNote('HURT');` results in 'HURT' + noteskin, if argument `texture` is null. HURTGOLDNOTE_assets doesn't exist for example, causing a crash in-game.

- Custom notes don't call `reloadNote()` to reload the texture in advance, meaning it could be overridden by the noteskin.
It doesn't load the texture in the chart editor, and has the potential to crash in-game.

- It is often impossible to determine the suggested path for a missing texture by looking at the terminal logs.

## The solutions:

- Calling `reloadNote('HURT', 'NOTE_assets');` results in 'HURT' + 'NOTE_assets', which conveniently exists. 

- Setting the custom note prefix to the note's name and then `reloadNote()` results in 'LUANOTE_assets' if it exists for example, avoiding null object references and loading the texture in the chart editor.

- If it doesn't, the texture isn't loaded in the chart editor and the terminal suggests to change the texture to a more conventional name.

# **Accepting this PR** means:

- Any custom note with those arguments will appear in the chart editor with it's texture, rather than just looking like a generic note.

- Noteskins will not effect any custom notetypes like Hurt Notes in-game.

- The game doesn't crash every funkin' time you want to use a noteskin on an existing chart.

- If other bugs cause the game to crash, the terminal will display more detailed stacks that can help solve problems.

### DIFF
--- a/source/Note.hx
+++ b/source/Note.hx
@@ -159,14 +159,16 @@ class Note extends FlxSprite
 				case 'GF Sing':
 					gfNote = true;
 				case '':
-
+					//normal notes retain the noteskin texture
+				case 'Hey!':
+					// hey notes retain the noteskin texture
 				default:
 					#if MODS_ALLOWED
 					luaPrefix = value.split(" ")[0].toUpperCase();
-					if (FileSystem.exists(Paths.image(luaPrefix)))
+					if (Paths.image(luaPrefix + 'NOTE_assets')!= null)
 						reloadNote(luaPrefix, 'NOTE_assets');
 					else
-						trace('Rename $value texture to ${luaPrefix}NOTE_assets');
+						trace('Suggestion: rename $value texture to ${luaPrefix}NOTE_assets');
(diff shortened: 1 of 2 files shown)


rodney528 (approved):
OoO

Itz-Miles:
PR is more useful now! All custom notes load in the chart editor, XML crash prevention stops notes from loading animations that don't exist, Paths does some path tracing for null assets, and no breaking changes either!

rodney528 (approved):

PR #11793, opened 2023-01-26
Fix MP4Handler issue

haya3218:
Due to latest hxCodec updates, `MP4Handler` was renamed to `VideoHandler`, and other stuff, which was already being done in detail in #11720 

This fixes the need to reinstall hxCodec version 2.5.1 by specifically specifiying the version in `hmm.json`.

Please use the `version` json var(?) next time. Something could break without proper notice.

PR #11821, opened 2023-01-30
Allow hxCodec to work with other platforms (MacOS, Linux)

ghost:


KirbyKid256:
Okay. Let me explain this the best I can.
[PolybiusProxy made a comment](https://github.com/ShadowMario/FNF-PsychEngine/pull/11793#issuecomment-1406809140) mentioning the fact that Psych Engine is using an older version of hxCodec.

These builds still use hxCodec 2.5.1 since it's installing from hmm.json which specifically grabs 2.5.1 now. #11793
While it does support Linux (even though there's currently something preventing it from compiling), it does NOT support macOS. You need to use hxCodec 2.6.1 (current) which DOES support macOS.

Now, in 2.6.1, MP4Handler still exists but has been moved and renamed to VideoHandler. Therefore, in PlayState.hx, instead of importing from `vlc.MP4Handler`, you'll need to import from `hxcodec.VideoHandler`. Then, you need to replace any instance of MP4Handler with VideoHandler

Once that's done, then you can change the project XML (as you did) to `desktop || web`. Unfortunately, if you don't update hxCodec to 2.6.1, then it simply can't work on macOS.

Luisinhi010 (approved):

KirbyKid256:
So, I don't know how to fix the issues with Linux yet. You should ask PolybiusProxy for help with it. But until that's fixed, you can use this to ingore Linux: `<define name="VIDEOS_ALLOWED" if="desktop || web" unless="32bits || linux"/>`

ActualMandM:
Maybe it needs to grab from the GitHub repository instead? There's this commit which seemingly fixes the linker error.
https://github.com/polybiusproxy/hxCodec/commit/15d950b103603a976e0e2d5004c8717b4eecc945

BeastlyGabi:
marking as draft as I need to test to see if we can compile to linux with the git builds

PR #11905, opened 2023-02-12
Another fix for hxCodec's new update

NexIsDumb:
For all the stinky people that updated hxCodec
This should fix the problems with hxCodec's new update checking the version that is installed (So it will work for everyone without depending on their version)

Edit: Now it also fixes the 2.6.0 version: so now it's possible to use 2.5.1, 2.6.0 or 2.6.1

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -72,7 +72,8 @@ import sys.io.File;
 #end
 
 #if VIDEOS_ALLOWED
-import vlc.MP4Handler;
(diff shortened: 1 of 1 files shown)


NexIsDumb (commented):
Yeah, Ive compiled several times using every version of hxCodec and they work with these changes

moxie-coder (approved):

NexIsDumb:
Now that I notice, this is what @BeastlyGabi did in #11821 but mine includes every version instead of just 2.6.1

aaaproooo (approved):

PR #11965, opened 2023-02-22
Null object references involving vocals and insts are prevented, and songs without vocals can be loaded in the chart editor.

GO-Miles:
This pull request changes **_Paths.hx_** and **_ChartingState.hx_** to properly deal with vocals, insts, and new songs that don't have those.

## Paths:
Reverted to an older filepath allocation method that still works since the graphics/memory overhaul.
## ChartingState:
Checked to see if the song actually needs voices before Paths tries to load an audio file. (The variable's file's type can't be checked if the file doesn't exist prior to the reference)
___
Both of these are untested! Lemme know if anything doesn't work as I did this in the web editor.

### DIFF
--- a/source/Paths.hx
+++ b/source/Paths.hx
@@ -372,6 +372,7 @@ class Paths
 	}
 
 	public static var currentTrackedSounds:Map<String, Sound> = [];
+	
 	public static function returnSound(path:String, key:String, ?library:String) {
 		#if MODS_ALLOWED
 		var file:String = modsSounds(path, key);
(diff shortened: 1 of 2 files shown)


moxie-coder (approved):

Itz-Miles:
I will most likely be implementing a placeholder asset system like the way haxeflixel handles FlxSprites. Too many null object references that can easily be taken care of!

Before I merge unrelated stuff however, I suggest others to check commits in patch 1: https://github.com/ShadowMario/FNF-PsychEngine/pull/11811

MemeHovy:
aight

Rom4chek:
Banger

roma-perec-bp (approved):

aaaproooo (approved):

lunacynne (approved):

PR #12029, opened 2023-03-05
Character instances can be created anywhere, and bugs/crashes involving character assets can be prevented.

GO-Miles:
# Overview
This pull request changes _**class character's `new()` function**_ to load character assets from any library, which prevents missing textures, null object refs, and crashes.

## The bugs:

* Paths.hx generally determines what library to load assets from based on the current FlxState. This means a character in MainMenuState can only load its texture from preload, and preloading assets while switching states can cause stuttering and crashes.

## The solutions:

Passing an optional library argument to Paths makes it load the path from the desired library rather than from a library based on what directory its set to.

# Accepting this PR means:

* modders can create new character instances wherever they please.
* preloading assets in HTML5 is more accurate and less error-prone when specified.
* Paths almost never fails to find the character files, regardless of where said files are.

lemme know if there are bugs, some of these are untested!

### DIFF
--- a/source/Character.hx
+++ b/source/Character.hx
@@ -79,7 +79,7 @@ class Character extends FlxSprite
 	public var healthColorArray:Array<Int> = [255, 0, 0];
 
 	public static var DEFAULT_CHARACTER:String = 'bf'; //In case a character is missing, it will use BF on its place
-	public function new(x:Float, y:Float, ?character:String = 'bf', ?isPlayer:Bool = false)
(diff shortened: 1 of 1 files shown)


charlesisfeline (approved):

moxie-coder (approved):

TheKitBoi (approved):

aaaproooo (approved):

Luisinhi010 (approved):

DM-kun (approved):

HenryBATIM:
Out of curiosity, what does it mean when "character instances" are specified?
I have a feeling I may know, but I wanna make sure.

PR #12030, opened 2023-03-06
Update credits

superpowers04:
* Change my twitter to my bio(I don't use twitter that often)
* Adjust my credit reason a bit (thank for the credits btw)


moxie-coder (approved):
👍🏻

Raltyro:
This does not solve miscredited contributors (gabriela, etc.) /hj

NexIsDumb (approved):
At this point I'd suggest adding smth to update the credits without everytime making a pr, maybe like checking an online file, similar to what the game does for checking updates

NexIsDumb:
> At this point I'd suggest adding smth to update the credits without everytime making a pr, maybe like checking an online file, similar to what the game does for checking updates

The problem is that every credited person should be aware and okay with this and create the actual file, sooo

Or maybe shadowmario creating an online doc where he adds the credited ppl as editors, idk

PR #12119, opened 2023-03-18
User Interface Class

ghost:
NOTICE: this will break a *ton*, and I mean a *tooon* of lua scripts that make use of the HUD variables

This change moves User Interface-related objects to a separate class in order to clean up playstate, it also prevents other ui objects from being created if a certain `ClientPref` or variable is disabled on PlayState

### DIFF
--- a/source/objects/AchievementPopup.hx
+++ b/source/objects/AchievementPopup.hx
@@ -33,10 +33,11 @@ class AchievementPopup extends FlxSpriteGroup {
 		add(achievementText);
 		add(achievementIcon);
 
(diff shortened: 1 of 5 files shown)


moxie-coder (approved):
pog

Fearester2008:
> has conflicts and yeah I think I'd be better to keep modding in mind

You can add it in my engine?

MemeHovy:
> > has conflicts and yeah I think I'd be better to keep modding in mind
> 
> You can add it in my engine?

dude, their account is deleted

PR #12123, opened 2023-03-19
Fix strum centering

UmbratheUmbreon:
The strums are not actually centered, and can be properly centered by increasing STRUM_X by 6.5.

image

As you can see in this image, before the change, the innermost notes where offset by 13. By dividing that by 2 to get 6.5, we can add that to STRUM_X to offset the strums by 6.5 units, which results in the following:

image

Perfectly centered strums!

The method used to find this was simply to create a sprite in the center of the HUD, that is 10 pixels wide and 720 pixels tall, then do a bit of math to get the distance each note is from the sprite's edges (i.e. its x and its x + width). The code for the trace is below:

image

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -81,7 +81,7 @@ using StringTools;
 
 class PlayState extends MusicBeatState
 {
-	public static var STRUM_X = 42;
+	public static var STRUM_X = 48.5;
 	public static var STRUM_X_MIDDLESCROLL = -278;
 
 	public static var ratingStuff:Array<Dynamic> = [


Kn1ghtNight (approved):
they are centered now im so happy

aaaproooo (approved):

moxie-coder (approved):

MemeHovy:
hey no hard feelings right @UmbratheUmbreon?
btw sorry for screwing up so much on denpa engine, I wasn't sure on what to do and what to add and I just wanted to make sure the engine and code was as good as much as it could be

charlesisfeline (approved):

CaptainBaldi:
i never looked at the strums closely enough to tell but now that i know they arent actually centered i might
explode

NexIsDumb (approved):

PR #12152, opened 2023-03-24
Why can't I touch this?

rodney528:
Being able to mess with this in lua would be useful... also ik getPropertyFromClass can _supposedly_ access this but it doesn't work for some reason.


frantastic24:
I've been using setProperty on this var for months and it worked fine, are you sure you didn't mess up the code

BeastlyGabi:
lua can access private variables

PR #12159, opened 2023-03-26
no more deprecated warnings

lemz1:
changes/fixes: 

- objects.AchievementPopup.hx
    - fixed the deprecated warnings by using @:privateAcess and getting the defaultCameras like that

### DIFF
--- a/source/Main.hx
+++ b/source/Main.hx
@@ -141,7 +141,7 @@ class Main extends Sprite
 			}
 		}
 
-		errMsg += "\nUncaught Error: " + e.error + "\nPlease report this error to the GitHub page: https://github.com/ShadowMario/FNF-PsychEngine\n\n> Crash Handler written by: sqirra-rng";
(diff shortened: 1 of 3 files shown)


moxie-coder (approved):
the other changes seem useless tho

lemz1:
yeah i guess but i always hate it when people just send the crash report and you have no information whatsoever.
the biggest issue is when you dont know the version because the error could have a different line for example on version 0.4.2 it could be on line 422 and on version 0.6.2 it could be on line 600.

MemeHovy:
fair

NexIsDumb (approved):

Kn1ghtNight:
why would you do this
```xml
<haxedef name="no-deprecation-warnings" />
```
is right there

MemeHovy:
> i think the approach with haxedef is bad, because if someone uses a deprecaded function, he isn't going to know.

I mean, you can disable it for debug builds, which is what I did

lemz1:
> I mean, you can disable it for debug builds, which is what I did

I guess thats true, but i'll just keep this pr open, so that if shadowmario actually wants to remove the deprecaded warning, then he can just pull this pr.

Kn1ghtNight:
> i think the approach with haxedef is bad, because if someone uses a deprecaded function, he isn't going to know.

deprecated functions get flagged in the docs???? (well not like most fnf coders read documentation)

Kn1ghtNight:
but literally if they wouldnt read docs they would end up trying to fix it themselves or asking for help
(most likely will do the second option) and most people would just say to ignore it and that its fine.

PR #12226, opened 2023-04-06
fix of 25 min music length cap

TheLeerName:
music length and time was negative when music length was higher than 1521742ms (25:21.742)
idk maybe psych dont needed that but ok
with this fix music length cap will be 12173936ms (3:22:53.936), and if you crazy guys exceeds this limit then song length will be equals max music length lol

### DIFF
--- a/source/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/source/lime/_internal/backend/native/NativeAudioSource.hx
@@ -487,7 +487,9 @@ class NativeAudioSource
 			return length;
 		}
 
-		return Std.int(samples / parent.buffer.sampleRate * 1000) - parent.offset;
+		var thelength:Int = Std.int(samples / parent.buffer.sampleRate * 1000) - parent.offset;
+		if (thelength < 0) thelength = 12173936;
+		return thelength;
 	}
 
 	public function setLength(value:Int):Int
--- a/source/lime/_internal/backend/native/NativeAudioSource.hx
+++ b/source/lime/_internal/backend/native/NativeAudioSource.hx
@@ -136,7 +136,7 @@ class NativeAudioSource
 			}
 		}
 
-		samples = Std.int((dataLength * 8) / (parent.buffer.channels * parent.buffer.bitsPerSample));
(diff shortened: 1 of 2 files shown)


moxie-coder (approved):

KookerFoxYT:
man d&b fans would have a field day with this one

charlesisfeline (approved):
i forgot to approve oops

KookerFoxYT:
one thing
shouldn't this be able to be merged to the experimental branch (v0.7b) instead of the main one because 0.7 is coming out soon

TheLeerName:
i did preventing negative value cool swag

TheLeerName:
> one thing shouldn't this be able to be merged to the experimental branch (v0.7b) instead of the main one because 0.7 is coming out soon

when experimental will be pushed in main, i edit it

CaptainBaldi:
tbh this seems like something that should just be added to a mod and not the base engine

TheLeerName:
> tbh this seems like something that should just be added to a mod and not the base engine

why not, more features more fun

NeeEoo:
Fyi: iirc the limit with this is 50 minutes, since the length of the song sample array overflows.

`@:privateAccess FlxG.sound.music._sound.__buffer.length` (fun fact: times 4 == ram usage)

And also the limit depends on the sample rate.

The number of 50 minutes was from when i made the same kind of fix, but used int64 instead of haxe's Ints (which are int32)

Where the line looked like this

`samples = Int64.toInt((Int64.make(0, dataLength) * Int64.ofInt(8)) / Int64.ofInt(parent.buffer.channels * parent.buffer.bitsPerSample));`

Only way to make it support longer songs would be to rewrite the way the song is loaded into memory, like by using either multiple arrays, or making it return an int64 when it gets the length in c++

libauror-a (approved):

lunacynne (approved):

aaaproooo (approved):

hackx2 (approved):

jimmyjamescool:
Mfs finna chart fatal finale. A 40 minute song into a funny beep boop newgrounds rhythm game modification

jimmyjamescool (approved):

ShadowMario:
Unnecessary, i fear it could cause any weird issues

PR #12260, opened 2023-04-13
Automated VersionShit (i guess)

hackx2:
the title said it all.


charlesisfeline (approved):
doesn't actually affect performance but sure

Hackx2:
> this is useless and you created an unnecessary variable called 'array' just use the 'versionShitArray'

I wouldn't say it's completely useless it just helps minimise the use of versionShit.

TheoFoxxo (approved):
cool?

moxie-coder (approved):

FunkoDev (approved):

BeastlyGabi:
this would probably be better for multiple version texts

```hx
var versionNames:Array<String> = [
	"Friday Night Funkin' v" + Application.current.meta.get("version"),
	"Psych Engine v" + psychEngineVersion
];

for (i in 0...versionNames.length) {
	var versionTxt:FlxText = new FlxText(0, 0, 0, versionNames[i]);
	versionTxt.setFormat("VCR OSD Mono", 16, FlxColor.WHITE, LEFT, FlxTextBorderStyle.OUTLINE, FlxColor.BLACK);
	versionTxt.setPosition(12, FlxG.height - versionTxt.height - 22);
	add(versionTxt);
}
```

BeastlyGabi:
and also because this is unnecessary

image

BeastlyGabi:
and also why is this a dynamic

image

the versions numbers are also strings

Hackx2:
Oh I apologise for my stupidity.

Hackx2:
It's also dynamic because I planned to add an x and y to the array. It was postponed since I didn't have enough free time to implement it. I never got around to converting it back to a string.

Hackx2:
Thank you for making that comment; it really helped me recognise what I did wrong.

NexIsDumb:
Kinda useless AT LEAST for the version
It would have been usefull for stuff you would edit most of the times, but -fortunately- this only get changed when a new version comes
So youre basically adding useless lines for something it would even be faster without this

Hackx2:
somewhat, but I hate the numerous versionshit variables

NexIsDumb:
> somewhat, but I hate the numerous versionshit variables

Aren't they just 2?
And plus your update added (we can call them like this for better understanding) out from function variable and the MainMenuState.hx has a persisten update, so your update made the game a very little heavier on the ram (again, very little), yet heavier for something it was shorter ig

UmbratheUmbreon:
> And plus your update added (we can call them like this for better understanding) out-from-function variables and the MainMenuState.hx has a persistent update, so your update made the game a very little heavier (even not when youre not in the main menu because of the persistent shit) on the ram (again, very little), yet heavier for something it was shorter before ig

Persistent update is a bool to determine whether the state updates when a substate is open, not if it updates when unloaded.

In addition:
Regardless of method, the texts are being add()ed to the state, and since states are groups, they become explicitly linked to the group until you remove() them. As a result, the texts will be automatically destroy()ed when the state is destroy()ed, so under normal conditions, it should not cause any lingering memory usage unless for some reason FlxText doesnt properly dispose of and/or put() its variables. (which it does)

> Aren't they just 2?

If this were reworked, it could potentially be useful to add text to the bottom left of the screen as needed in any state, in addition to its use for the main menu, in a similar vein to that of addLuaDebugText in PlayState

NexIsDumb:
> If this were reworked, it could potentially be useful to add text to the bottom left of the screen as needed in any state, in addition to its use for the main menu, in a similar vein to that of addLuaDebugText in PlayState

Yeah but why would you need this?
I mean, as you said addLuaDebugText exists, and if you want to add any other texts in any other state like that its just 2/3 lines

Hackx2:
> Kinda useless AT LEAST for the version It would have been usefull for stuff you would edit most of the times, but -fortunately- this only get changed when a new version comes So youre basically adding useless lines for something it would even be faster without this

Not necessarily, because some people fork psych engine and use it as the foundation for their own engine.

Hackx2:
**True.**

When you say txt file do you mean something like the credits txt file?

NexIsDumb:
> **True.**
> 
> When you say txt file do you mean something like the credits txt file?
> 
> 

Yeah something like that, why not
Better not as json

UmbratheUmbreon:
> And plus no, persistent update it is used as I said
What you meant for substates is persistentDraw

PersistentUpdate and PersistentDraw go hand in hand, in the situation of substates. Swapping states _unloads the state_, and as such _persistentUpdate and Draw **do not have any effect when the state is changed to another**_.

The only way they could affect performance in this instance is if HaxeFlixel did not unload states when swapping them, and just kept stacking them. HaxeFlixel does not do this.

Quoted directly from the HaxeFlixel api:

Screenshot_20230419-084658~2

Hackx2:
> > **True.**
> > When you say txt file do you mean something like the credits txt file?
> 
> Yeah something like that, why not
> 
> Better txt than json for these kind of stuff
> 
> But anyways why are you committing into main? 0.7 is getting worked in the experimental branch

I forgot to fork the experimental branch.

https://github.com/ShadowMario/FNF-PsychEngine/pull/12287

NexIsDumb:
> > > **True.**
> > > When you say txt file do you mean something like the credits txt file?
> > 
> > 
> > Yeah something like that, why not
> > Better txt than json for these kind of stuff
> > But anyways why are you committing into main? 0.7 is getting worked in the experimental branch
> 
> I forgot to fork the experimental branch.
> 
> #12287

k

rodney528 (approved):

PR #12307, opened 2023-04-23
Json Options for Modpacks - COMPLETED

NexIsDumb:
~THIS IS STILL UNDER WORK!!~  **COMPLETED!! - LOOK FROM RIGHT BEFORE THE SCREENSHOTS FOR MORE INFOS**
Yeah, the title says all, I started working on this few days ago after some ideas
Here's everything (I'll prob add more stuff in the future to the list):
- [x] Add a function in Paths that returns active mods folders and simplified _pushGlobalMods()_ (I will prob simplify other source parts in the future)
- [x] Add the options folder with a readme and code to read the json and add the option
- [x] Let the program know when the option is from json (to call the variable value from a saved map prob)
- [x] Add functions on Lua for getting the saved prefs for json options (and prob for normal ones too, to avoid future retro-compatibilty break)
- [x] Make the saved options json saved prefs load the default value if never even saved that option inside the saved map
- [x] Add a substate for deleting saved prefs of modpacks
- [x] Prob more shits that I cannot remember rn
- [x] Optimizations check

If you have feedbacks, please tell them
(If this gets added would be nice if credited)

Edit: I added some callbacks that are mainly for the json options but also should have fixed eventual future compatibility breaks when saving / getting a pref from ClientPrefs

**YOU CAN BUILD THE OPTION (Using a json) AS YOU WANT JUST LIKE IN THE SOURCE**
Dont mind the options I made for the screenshots theyre ugly LOL
You can get the saved prefs on lua trough some callbacks I added (Read the redme inside the options folder)

Example of how you can make an option in a json (some variables have their default values, specially the name will have the file name if the variable "name" is null) - (the minimum that you can add is just the variable, THATS THE MAIN THING!)
You’ll have to place these json inside the options folder and then you can continue the path based on the menu where you want to place them (using their class names, ex: VisualsUISubState, GameplaySettingsSubState etc..)
The field system for adding the function, is dynamic, means you won’t need to edit anything about this json system if you add for example a new variable inside the Option class or if you, instead remove one.

```json
{
    "name": "Game Mechanics",
    "variable": "mechanic"
}
```

or

```json
{
    "name": "Mechanics Difficulties: ",
    "description": "How difficult should the mechanics be.",
    "variable": "mechDiff",
    "type": "string",
    "options": ["Easy", "Normal", "Hard"],
    "defaultValue": "Normal"
}
```

etc..

If this is going to get added, youre totally free to change and improve anything

Screenshots:
image
image
image
image
image
image

**When you remove the modpack from the mods folder yet there are still options saves in the memory:**
image

**No options saves found:**
image

Prob the readme is too long but we'll see Ig

### DIFF
--- a/example_mods/options/readme.txt
+++ b/example_mods/options/readme.txt
@@ -8,11 +8,16 @@ If the option name variable is null, the game will use json file name.
 
 For getting the saved option data on lua, you can use: getOptionSave([Variable Name], [If it's a json Option], [The name of the Mod's Option (Mod's Folder)])
 For setting its data: setOptionSave([Variable Name], [Value you'd like to set], [If it's a json Option], [The name of the Mod's Option (Mod's Folder)])
+To prevent unloaded saved options prefs: loadJsonOptions([If you'd like to include the main global mods folder], [The names of the Mods' Options' (Mods' Folder) as array])
 Note that:
 [If it's a json Option] as default is false (This means you can use these for Normal Options too)
 [The name of the Mod's Option (Mod's Folder)] as default is where the whole mod is stored, if you change this, you have the possibility to check other mods prefs too.
+[If you'd like to include the main global mods folder] as default is true
 Examples:
 getOptionSave('downScroll')
 getOptionSave('mechanic', true)
 getOptionSave('mechanic', true, 'daModFolder')
 setOptionSave('mechanic', false, true)
+loadJsonOptions(true, {'daModFolder', 'anotherModFolder'})
+loadJsonOptions(false)
+loadJsonOptions()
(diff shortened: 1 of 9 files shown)


charlesisfeline (approved):

moxie-coder (approved):

NexIsDumb:
With these last added lua callbacks I should have also fixed eventual future compatibility breaks when saving / getting a pref from ClientPrefs

NexIsDumb:
Clarification about loadJsonOptions (I'll include and summarize for stinky modders this the next time I'll update the readme): Basically, the variable and saved prefs about an option gets stored a save map in ClientPrefs mainly only when the player enters an option menu where specific options where loaded.
What if the player never loaded the json option in the option menu? The prefs got never loaded or saved, and so the get / set callbacks about the options json prefs will never find the specified unloaded pref.
To prevent this I added loadJsonOptions, that generally loads and save the unloaded options of specified mods (or also from the global folder too)
So we can say that this is an half init callback

In the next commit I'll also add an easy way to call ClientPrefs.saveSettings() on lua so modders can use it mainly after setting an option pref

NexIsDumb:
Ig its done lmao

Joalor64 (approved):
ayo?? 👍

NexIsDumb:
> ayo?? 👍

Got more ideas to improve it a little bit now

Temmiekat (approved):

ManMaster707 (approved):

PR #12313, opened 2023-04-23
Simplified Paths.hx

Joalor64:
It's pretty self-explanatory.

### DIFF
--- a/source/Paths.hx
+++ b/source/Paths.hx
@@ -99,7 +99,6 @@ class Paths
 		for (key in currentTrackedSounds.keys()) {
 			if (!localTrackedAssets.contains(key)
 			&& !dumpExclusions.contains(key) && key != null) {
-				//trace('test: ' + dumpExclusions, key);
 				Assets.cache.clear(key);
 				currentTrackedSounds.remove(key);
 			}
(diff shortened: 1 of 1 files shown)


moxie-coder (commented):
@Joalor64GH the main reason those are two separate things is because it'll return a different type if it is done like this, that's the main pet peeve I have with haxe but that's how it works though, believe me, I've tried before

MemeHovy:
unless you add the return type however

source/Paths.hx
```diff
@@ -139,54 +136,37 @@ class Paths
 	}
 
 	static public function getLibraryPath(file:String, library = "preload")
-	{
 		return if (library == "preload" || library == "default") getPreloadPath(file); else getLibraryPathForce(file, library);
-	}
 
 	inline static function getLibraryPathForce(file:String, library:String, ?level:String)
```
MemeHovy, line 141: make sure to add the return types for some of these, otherwise it'll throw errors, cuz the compiler will auto assign an type to it

moxie-coder (changes requested):

moxie-coder (changes requested):
returnGraphic doesn't have a type assigned to it so it'll be automatically assigned, likely to cause errors

CaptainBaldi:
Most of these changes are completely useless

MemeHovy:
> Most of these changes are completely useless

that aswell

KookerFoxYT:
> Most of these changes are completely useless

agreed

KookerFoxYT:
i mean like, it isnt that much more simplified than what one can do to simplify ClientPrefs

PR #12320, opened 2023-04-26
added callOnHscript

Blossomical:
added that since runHaxeCode has a chance of crashing when its used alot on functions that are constantly ran like "onUpdate"


elchango69:
im eating jolly ranchers from my ass

Cherif107:
closed because sm already added it

PR #12335, opened 2023-04-28
A bit of clarification in ModsMenuState

moxie-coder:
might be useless, granted that the back key can be seen in controls, although just having it say "BACK" instead of "ESCAPE" or "BACKSPACE" might not be clear to _some_ people, so I made it just slightly more clear to some people.

(And also removed old code that wasn't needed, just because you can.)

Also the extra var was needed since it wouldn't compile otherwise.

Fun Fact: this is an old PR I forgot about (as in, I forgot to push to github)


jimmyjamescool:
pp balls

jimmyjamescool (approved):

roma-perec-bp (approved):

Issue #12338, opened 2023-04-29
Updated dependencies causing errors


RowanSkie:
### Describe your bug here.

I've went and run ```haxelib run hmm install``` and after flixel and flixel-addons updated, I get these problems.

What's the recommended versions for those that updated? Can we get an updated hmm for this?

### Command Prompt/Terminal logs (if existing)

```bash
PS C:\Users\xterm\Documents\GitHub\FNF-PsychEngine> haxelib run hmm install
haxelib --never install lime
You already have lime version 8.0.1 installed
haxelib --never install openfl
You already have openfl version 9.2.1 installed
haxelib --never install flixel
Downloading flixel-5,3,0.zip...
Download complete: 0.24KB in 3.1s (0KB/s)
Download complete: 1.15MB in 1.8s (606.2KB/s)
Installing flixel...
  Current version is now 5.3.0
Done
haxelib --never install flixel-addons
Downloading flixel-addons-3,1,0.zip...
Download complete: 0.25KB in 1s (0.2KB/s)
Download complete: 417.66KB in 2.3s (176.3KB/s)
Installing flixel-addons...
  Current version is now 3.1.0
Done
haxelib --never install flixel-tools
You already have flixel-tools version 1.5.1 installed
haxelib --never install flixel-ui
You already have flixel-ui version 2.5.0 installed
haxelib --never install hxcpp
You already have hxcpp version 4.3.2 installed
haxelib --never install tjson
You already have tjson version 1.4.0 installed
haxelib --never install hxjsonast
You already have hxjsonast version 1.1.0 installed
haxelib --never install hxCodec 2.5.1
You already have hxCodec version 2.5.1 installed
haxelib --never install hscript
You already have hscript version 2.5.0 installed
haxelib --never install hxcpp-debug-server
You already have hxcpp-debug-server version 1.2.4 installed
haxelib --never git discord_rpc https://github.com/Aidan63/linc_discord-rpc master
You already have discord_rpc version git installed.
Updating discord_rpc version git ...
discord_rpc was updated
Library discord_rpc current version is now git
Installing dependency hxcpp 4.3.2
You already have hxcpp version 4.3.2 installed
haxelib --never git hscript-ex https://github.com/ianharrigan/hscript-ex master
You already have hscript-ex version git installed.
Updating hscript-ex version git ...
hscript-ex was updated
Library hscript-ex current version is now git
[...]

PR #12366, opened 2023-05-04
Lock library versions, pending Haxe 4.3.1 fixes

EliteMasterEric:
There is an issue where Psych Engine does not build with the latest versions of Haxe and Flixel, as seen with issues like 
#12338.

In order to build Psych right now, users are required to do the following:

- Install Haxe 4.2.5 and use it instead of Haxe 4.3.1.
- Install specific versions of dependent haxelibs

This fix resolves the latter, by telling `hmm` to install and use specific versions of libraries, rather than assuming that the latest Haxelib version will always work.

Later on, the `hmm` file should be edited to point to the latest version of Flixel and Flixel-Addons, then Psych Engine's bugs should be fixed to work with these, so that users can finally move to Haxe 4.3.1.

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -1,49 +1,43 @@
 {
   "dependencies": [
     {
-      "name": "lime",
-      "type": "haxelib",
-      "version": null
-    },
-    {
-      "name": "openfl",
-      "type": "haxelib",
-      "version": null
+      "name": "discord_rpc",
+      "type": "git",
+      "dir": null,
+      "ref": "2d83fa8",
+      "url": "https://github.com/Aidan63/linc_discord-rpc"
     },
     {
       "name": "flixel",
       "type": "haxelib",
-      "version": null
+      "version": "4.11.0"
     },
     {
       "name": "flixel-addons",
       "type": "haxelib",
-      "version": null
+      "version": "2.11.0"
     },
     {
       "name": "flixel-tools",
       "type": "haxelib",
-      "version": null
+      "version": "1.5.1"
     },
     {
       "name": "flixel-ui",
       "type": "haxelib",
-      "version": null
+      "version": "2.5.0"
     },
     {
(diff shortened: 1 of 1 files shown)


moxie-coder (approved):

jimmyjamescool:
22 dicks up fortnite asscheeks

jimmyjamescool (approved):

AltronMaxX:
I think that is not a problem of Haxe 4.3.1, it's only HaxeFlixel issue

NexIsDumb:
> I think that is not a problem of Haxe 4.3.1, it's only HaxeFlixel issue

this fixes also flixel stuff wdym

NexIsDumb (approved):

ender69funk:
It's not helping this fuckin hmm

SlickFromMars:
You might want to include the tidbit about only using Haxe 4.2.5 on the ``README.md``

DragShot:
Phew, I see I did well in coming to pay a visit over here! I've locked the libraries installed in my PC, and these are the versions that work for me:

- haxe `4.2.5`
- lime `8.0.1`
- openfl `9.2.1`
- flixel `5.2.2`
- flixel-addons `3.0.2`
- flixel-tools `1.5.1`
- flixel-ui `2.5.0`
- hxcpp `4.2.1`
- tjson `1.4.0`
- hxjsonast `1.1.0`
- hxCodec `2.5.1`
- hscript `2.5.0`
- hxcpp-debug-server `1.2.4`

I guess these are the newest versions I can recommend for compiling Psych Engine at the moment.

Stilic:
I did [a PR](https://github.com/ShadowMario/FNF-PsychEngine/pull/12561) that fixes the compilation issues (I successfully compiled it on Arch Linux with the latest Haxe version).

Capture du 2023-06-10 08-45-35

PR #12468, opened 2023-05-24
Re-use popupScore sprites instead of constantly destroying and re-adding.

MeguminBOT:
Should improve overall performance of the game, less likely to cause frame stuttering, since the game doesn't constantly have to remove and add the sprites everytime a note is hit. 

There may be a better method to do this and in that case I'm all ears, but this is atleast better than the current way the sprites are handled.

### DIFF
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -2548,16 +2548,29 @@ class PlayState extends MusicBeatState
 		rating.x += ClientPrefs.data.comboOffset[0];
 		rating.y -= ClientPrefs.data.comboOffset[1];
 
-		var comboSpr:FlxSprite = new FlxSprite().loadGraphic(Paths.image(pixelShitPart1 + 'combo' + pixelShitPart2));
+		var comboSprGroup:FlxSpriteGroup = new FlxSpriteGroup();
+		var comboSpr:FlxSprite;
+		var comboSprX:Float = coolText.x;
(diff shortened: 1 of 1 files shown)


moxie-coder (approved):

ghost (approved):
not bad. just good

MemeHovy:
> not bad. just good

true

NeeEoo:
This wouldnt work since the groups doesnt remember the content the next time you hit a note, you would need to make the groups outside of the function (such as in a field) or in create.

MemeHovy:
> This wouldnt work since the groups doesnt remember the content the next time you hit a note, you would need to make the groups outside of the function (such as in a field) or in create.

hmm, I have to agree with him here, what could be done is saving the content in another group or something

BeastlyGhost:
the group needs a way to know which sprite comes first by sorting the groups with a value that represents the songPosition, e.g:

```hx
class ComboSprite extends FlxSprite {
	public var depth:Float = 0;
	public static inline function sortByDepth(i:Int, a:ComboSprite, b:ComboSprite):Int
		return a.depth > b.depth ? -i : i;
}

// then when creating the sprite, set the depth of the sprite to the song position, like this
rating.depth = -Conductor.songPosition;
```

rodney528 (approved):

ShadowMario:
Draft has not been updated since December 2023, closing this

PR #12617, opened 2023-06-24
Improve compatibility with Flixel 5.3.0 and Add events "Set Cam Zoom" and "Flash"

QuetzalcoutlDev:
Fixes in the PlayState.hx and FunkinLua.hx files to allow compilation without using the LUA_ALLOWED and MODS_ALLOWED variables

Some commits are for my confusion, the changes I have applied are to allow Psych Engine compatibility with Flixel 5.3.0, allow compiling the Engine without Lua compatibility and the mods folder


moxie-coder (approved):

jimmyjamescool (approved):

CaptainBaldi:
For things like the `FlxSound` fix it would probably be more optimal to check which version of Flixel the user has and import the library based on that. A lot of people don't know to update their libraries which would still cause more issues to be made here.

QuetzalcoutlDev:
@CaptainBaldi Well, that's true, that didn't cross my mind, so I'll fix that.

What I thought is that those who install Haxeflixel for the first time download the latest version and end up with the problems that I tried to solve

JokeBambi (approved):
nice

PR #12654, opened 2023-06-30
Add defaultAntialiasing to FlxSprite without Flixel 5

teotm:
Flixel 5 added added `defaultAntialiasing` variable to `FlxSprite`s. Some peeps might not use Flixel 5 for whatev reason so yeh.

### DIFF
--- a/source/flixel/FlxSprite.hx
+++ b/source/flixel/FlxSprite.hx
@@ -1,4 +1,4 @@
-package flixel.old;
+package flixel;
 
 import flash.display.BitmapData;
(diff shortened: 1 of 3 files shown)


moxie-coder (approved):
interesting

NexIsDumb:
Wait
At this point if you're making a fork you could do directly
```haxe
public var antialiasing(default, set):Bool = ClientPrefs.data.antialiasing;
```
And editing the ClientPrefs code in order to make the most of this thing

PR #12685, opened 2023-07-05
Fix macro errors on Haxe 4.3.0+

libauror-a:
quick patch for haxe versions 4.3.0+


MemeHovy:
they're already fixed, look in project.xml

ThePlank:
> they're already fixed, look in project.xml

oops didnt see that

PR #12708, opened 2023-07-08
fixed the ordinal number formatting in ChartingState.hx

ms-kaylaa:
it sort of bugged me so i tried to fix it so instead of "32th" it'd say "32nd"
its nowhere near optimal but i tried


moxie-coder (approved):

TheLeerName (approved):
no way it is the spelling update

rodney528 (approved):

PR #12732, opened 2023-07-13
Shader coords fix

Sword352:
Just a pull request that applies the shader coords bug fix globally instead of only on PlayState, lua scripts and such. Also contains a little optimization to PlayState


moxie-coder (approved):
slay

PR #12733, opened 2023-07-13
include classes which not used in game

TheLeerName:
hscript: 📈

for example u cant `addHaxeLibrary('FlxGlitchEffect', 'flixel.addons.effects.chainable')` if source code not used this class


MemeHovy:
couldn't you just add them here instead?
Untitled50

CaptainBaldi:
> couldn't you just add them here instead? Untitled50

considering theres probably hundreds of libraries Psych Engine doesn't use its probably more efficient to do this :shrug:

MemeHovy:
I mean, all the macro does as far as I know is just import classes in a more sophisticated way, that hscript I think can use?
but this doesn't seem as far fletched as just adding them into the interpreter

TheLeerName:
for example u cant import FlxBackdrop if source code not used this class

MemeHovy:
> for example u cant `addHaxeLibrary('FlxGlitchEffect', 'flixel.addons.effects.chainable')` if source code not used this class

okay but like, why not just add them to the interpreter? It’d likely be the same thing as what you did

TheLeerName:
> okay but like, why not just add them to the interpreter if you did use source? It’d likely be the same thing as what you did

bro like its hundreds of lines, and some classes can have same name but other package

MemeHovy:
> > okay but like, why not just add them to the interpreter if you did use source? It’d likely be the same thing as what you did
> 
> bro like its hundreds of lines, and some classes can have same name but other package

hmm, I cannot argue with that

moxie-coder (approved):

TheLeerName:
#12848

PR #12772, opened 2023-07-21
fix retry softlock  😁😁👍

Snoni:
if you uh press enter on retry it plays the animation and never restarts the song  so in this pull request it fixes it 😁👍


MemeHovy:
did you even test this to make sure it actually fixes it

Snoni:
> did you even test this to make sure it actually fixes it

yes 😁👍

moxie-coder (approved):
alright bet

PR #12850, opened 2023-07-25
using lua functions without onCreate callback

TheLeerName:
lua: 📈

for example, u can use this line as .lua file, and it will work:
```lua
setProperty('skipArrowStartTween', true)
```

all pr just about do callbacks BEFORE loading lua file (game doing it AFTER)

### DIFF
--- a/source/psychlua/FunkinLua.hx
+++ b/source/psychlua/FunkinLua.hx
@@ -71,27 +71,8 @@ class FunkinLua {
 		//LuaL.dostring(lua, CLENSE);
 
 		this.scriptName = scriptName;
-		try{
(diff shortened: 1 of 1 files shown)


moxie-coder (approved):
hmm, this seems nice

MemeFlavor (approved):
hell yeah

CaptainBaldi:
what exactly does this do? im kinda confused

TheLeerName:
> what exactly does this do? im kinda confused

earlier u can use script only like this:
```lua
function onCreate()
    setProperty('skipArrowStartTween', true)
end
```

now u can use this script too:
```lua
setProperty('skipArrowStartTween', true)
```

doom699:
loving this

PR #12896, opened 2023-07-28
make changing bpm a setter instead of an entire function

Rudyrue:
this allows for lerping/tweening the bpm which would be really helpful on sliding bpms, like in too slow from sonic.exe

also it's just better to use overall cuz it's just setting a variable lmfa

### DIFF
--- a/source/backend/Conductor.hx
+++ b/source/backend/Conductor.hx
@@ -14,7 +14,7 @@ typedef BPMChangeEvent =
 
 class Conductor
 {
-	public static var bpm:Float = 100;
+	public static var bpm(default, set):Float = 100;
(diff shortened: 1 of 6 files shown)


crowplexus (approved):
should be like that from the start

moxie-coder (approved):

Sword352 (approved):
goATED 🗣🔥🫵💯

DetectiveBaldi (approved):
tbh a easy way to tween bpm would be really useful

charlesisfeline (approved):
yes

NexIsDumb (approved):
GODDAMIT finally

PR #12903, opened 2023-07-28
Rewrite SScript Implementation

ghost:
This PR rewrites most of the SScript implementation.

- Ditched `doString`, HScript uses now `new` which is better optimized.
- Due to `doString` change, `runHaxeCode` can now use files. `getTextFromFile` is unnecessary.
- `runHaxeCode` now can return values without a function.
- Scripts now gets cleared from memory completely using `destroy()`.
- `global` is used to check for conflicting scripts.
- Rewrote `callOnHScript`, which fixes not being able to call the same function from multiple scripts.
- Replaced `interpName` with `origin` which is used for Lua traces.
- `addHaxeLibrary` now doesn't initialize Haxe scripts for its Lua script, so you can use it after or before `runHaxeCode`.

### DIFF
--- a/source/psychlua/FunkinLua.hx
+++ b/source/psychlua/FunkinLua.hx
@@ -249,7 +249,7 @@ class FunkinLua {
 				args = [];
 			}
 
-			var foundScript:String = findLuaScript(luaFile);
+			var foundScript:String = findScript(luaFile);
 			if(foundScript != null)
 				for (luaInstance in game.luaArray)
 					if(luaInstance.scriptName == foundScript)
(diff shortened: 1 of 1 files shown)


charlesisfeline (approved):
better

Rudyrue (changes requested):
i think you looked over a line in playstate line 823

TheWorldMachinima:
> i think you looked over a line in playstate line 823

There is nothing in line 823.

Rudyrue:
image

?

TheWorldMachinima:
> image
> 
> ?

Are you sure you're using my fork?

Rudyrue:
that's odd it's not on your fork

ig it's just a problem on my end nvm
i fixed it anyways

Joalor64 (approved):

moxie-coder (approved):

ghost (approved):
bros cooking heat?

PR #12922, opened 2023-07-29
Rewrite SScript Implementation

ghost:
This PR rewrites most of the SScript implementation.
(Resolved the conflicts)

- Ditched `doString`, HScript uses now `new` which is better optimized.
- Due to `doString` change, `runHaxeCode` can now use files. `getTextFromFile` is unnecessary.
- `runHaxeCode` now can return values without a function.
- Scripts now gets cleared from memory completely using `destroy()`.
- `global` is used to check for conflicting scripts.
- Rewrote `callOnHScript`, which fixes not being able to call the same function from multiple scripts.
- Replaced `interpName` with `origin` which is used for Lua traces.
- `addHaxeLibrary` now doesn't initialize Haxe scripts for its Lua script, so you can use it after or before `runHaxeCode`.

### DIFF
--- a/source/psychlua/FunkinLua.hx
+++ b/source/psychlua/FunkinLua.hx
@@ -1511,6 +1511,9 @@ class FunkinLua {
 		if(hscript != null)
 		{
 			hscript.active = false;
+			#if (SScript >= "3.0.3")
+			hscript.destroy();
+			#end
 			hscript = null;
 		}
 		#end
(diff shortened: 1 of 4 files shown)


moxie-coder (approved):

ghost (approved):

ACoolioDude (approved):
Yes

Intancote (approved):

Joalor64 (approved):

lunacynne (approved):

charlesisfeline (approved):

frantastic24:
this seems to have broken scripts that use runHaxeCode more than once?

PR #12078, opened 2023-03-12
Basic Improvements to Lua's setHealthBarColors

DetectiveBaldi:
Currently, `setHealthBarColors` is sufficient enough, however if you are trying to change one side of the healthbar's colors, it's not too great at that.
Currently, running the code `setHealthBarColors(nil, 'FF0000')` in your Lua script presents you with this.

before
Due to the way this function is coded, trying to change one side of the healthbar's colors with it isn't ideal.

This PR fixes that, if one argument of the function is nil/empty, it will simply use the current health bar color for the switch.
Now, running the same example mentioned above provides you with this.

after
Thus making changing one half of the health bar's colors much simpler.

Reopen of #12074

PR #12924, opened 2023-07-29
HealthBar/General Bar Improvements

DetectiveBaldi:
Pretty much #12078 but working on newer versions with more features.

1. Renamed `HealthBar` class to `Bar`, cause yknow, not every instance of it IS a healthbar.

2. You can now set the health limit to any value using Lua and HScript.
The previous limit was 2, and as far as I know, couldn't be properly adjusted, due to `PlayState` having a line of code that limits how far the `health` variable can increase to before getting set back to 2.

Now, that statement is based on the healthbar's `bounds.max` property:

image

So now you can use `setProperty('healthBar.bounds.max', 5)` or `runHaxeCode('game.healthBar.bounds.max = 5;')` in order to increase the health limit.

3. Improvements to `setColors` / Lua's `setHealthBarColors`
On the `setColors` side, I just added some simple null checks to prevent the `leftBar/rightBar` colors from being set if the `left/right` parameters are null.

For `setHealthBarColors`, now you can set one side of the bar and leave the other side the same color, for example, running the code `setHealthBarColors('red', nil)` in current builds produces:

before

Now, it produces:

after

hoping this doesnt break anythinggg :pray:


charlesisfeline (approved):
the old pr may be good but this is infinitely better

moxie-coder (approved):

Intancote (approved):

NexIsDumb (approved):
Hell yeah

Issue #12954, opened 2023-07-31
shadowmario you broke hscript again :sob:


CaptainBaldi:
### Describe your bug here.

I can only use `runHaxeCode` once per script it seems. Every instance after that simply doesn't run.
For example I have a `runHaxeCode` that runs on create to replace the cursor graphic, I also have another that removes the cursor graphic on destroy.
The one on create works because it's the first instance to run, the other one doesn't
0.7.1 hotfix hotfix :pensive:

### Are you modding a build from source or with Lua?

Lua

### What is your build target?

Windows

### Did you edit anything in this build? If so, mention or summarize your changes.

no

ShadowMario:
Technically this wasn't my fault but i get what is the problem lol

CaptainBaldi:
oh well, at least its acknowledged

TheLeerName:
[btw temporary fix as script!!!!!!](https://drive.google.com/file/d/1e1BnGQcoL5ypAj0MA7Upg8iC-XLZ5szj)

PR #13017, opened 2023-08-06
rewrite of HScript.hx (fix of runHaxeCode and etc)

### LINKED ISSUE #13012: RunHaxeCode can only run once
### Describe your bug here.

Trying to run more than 1 `runHaxeCode` in a lua file, will not work past 0.6.3 versions.
This is because 0.6.3 initialized HScript and then ran the code, while the newer versions initialize HScript and run the code in the same function. Normally this wouldn't be a problem if the function didn't look like this:
`	public static function initHaxeModuleCode(parent:FunkinLua, code:String)`
`	{`
`		#if (SScript >= "3.0.0")`
`		if(parent.hscript == null)`
`		{`
`			trace('initializing haxe interp for: ${parent.scriptName}');`
`			parent.hscript = new HScript(parent, code);`
`		}`
`		#end`
`	}`
because once it runs `parent.hscript` isn't `null` anymore and then the code will never be able to be ran using `runHaxeCode` in LUA.

### Command Prompt/Terminal logs (if existing)

_No response_

### Are you modding a build from source or with Lua?

Lua

### What is your build target?

Windows

### Did you edit anything in this build? If so, mention or summarize your changes.

No.

TheLeerName:
code in runHaxeCode runs only if you did it one time, if call it again, it ignores code, this pr fixes that problem
resolves #13012, #12954
u need use `public var examplevar = 0;` to use `examplevar` in another calls of runHaxeCode tho

- **addHaxeLibrary can add enums now**
- **rewrited some error messages**

1. for example (new):
```
ERROR ON LOADING - mods/scripts/sdf.hx:2: Unknown variable: pig
ERROR (onCreate) - mods/scripts/sdf.hx:2: Unknown variable: pig
```
2. also "initialized sscript interp successfully" now traces even if onCreate function doesnt exist
3. also error messages from hscript now traces to console

- **rewrited some code in HScript.hx (optimization?!??!?)**
- **added new coolie feature for 2nd argument of runHaxeCode**
1. for example (new feature):
```lua
runHaxeCode("debugPrint(songPercent);", {'game'})
```
2. the same as:
```lua
runHaxeCode("debugPrint(game.songPercent);")
```

### DIFF
--- a/source/psychlua/FunkinLua.hx
+++ b/source/psychlua/FunkinLua.hx
@@ -1450,7 +1450,7 @@ class FunkinLua {
 			trace(e);
 			return;
 		}
-		trace('lua file loaded succesfully:' + scriptName);
+		trace('lua file loaded successfully:' + scriptName);
 
 		call('onCreate', []);
 		#end
@@ -1485,7 +1485,7 @@ class FunkinLua {
 			// Checks if it's not successful, then show a error.
 			if (status != Lua.LUA_OK) {
 				var error:String = getErrorMessage(status);
-				luaTrace("ERROR (" + func + "): " + error, false, false, FlxColor.RED);
+				luaTrace("ERROR (" + func + ") - " + error, false, false, FlxColor.RED);
 				return Function_Continue;
 			}
 
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -5,6 +5,11 @@ import objects.Character;
 import psychlua.FunkinLua;
 import psychlua.CustomSubstate;
 
+#if sys
+import sys.FileSystem;
+import sys.io.File;
+#end
+
 #if (HSCRIPT_ALLOWED && SScript >= "3.0.0")
 import tea.SScript;
 class HScript extends SScript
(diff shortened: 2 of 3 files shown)


moxie-coder (approved):

TheLeerName:
fuck you workflows never used you

mcagabe19:
@MemeHovy think two time before approving pls

TheLeerName:
so, we need to wait new update of rewritten classes
u can merge this, but scripts with classes just will not work as earlier (scripts without classes will work tho)

ghost (approved):
tested works fine

MemeHovy:
> @MemeHovy think two times before approving pls

alright

rodney528 (approved):

Meloomazy (approved):

rodney528 (approved):

ghost (changes requested):
can't you just add ``#if (SScript < "5.0.0")``? fully removing class support is pointless

mcagabe19:
> > can't you just add `#if (SScript < "5.0.0")`? fully removing class support is pointless
> 
> class support doesnt work in 5.0.x and older so

I still use class support in my repo and works great at older version of SScripts

TheLeerName:
i can but everyone uses newest versions of libs, so its will be kinda useless

RodneyAnImaginativePerson:
> i can but everyone uses newest versions of libs, so its will be kinda useless

How so?

TheLeerName:
now it useless cuz SScript is not exists
#13304

PR #13083, opened 2023-08-13
Always use latest versions

ghost:
so we not gonna get god damn ``set-output`` warnings


mcagabe19:
> probably a duplicate of #11340 but whatever the pr i mentioned is 8 months old so i'll let it pass

and it's closrd by ShadowMario

PR #13121, opened 2023-08-18
Hide Misses Text while "Instakill on Miss" Gameplay Modifier is enabled

crowplexus:
self explanatory


moxie-coder (approved):
makes sense

PR #13141, opened 2023-08-20
Icon workaround fix on Linux

nubz4lif:
linux doesn't have app icons (window/taskbar icons)

so instead, add icon.png to the main folder, and set the icon manually in Main.hx

i tried using Assets.getBitmapData too, but it would result in the icon being displayed in BGR instead of RGB


moxie-coder (approved):
don't use linux, but seems great

ghost (approved):
mmmm

PR #13243, opened 2023-09-05
Fixes for newer haxe versions and compile scripts

crowplexus:
thanks @Rudyrue for the batch script cuz I'm stupid yeye

* Haxe 4.3 and higher versions now compile properly
* We included new scripts to compile the game easily


Rudyrue:
yipe............................................................................................................

Stilic:
Why isn't the setup script parsing the hmm.json file instead of a separate deps file?
This would allow the usage of the old method for people that have this method working for them.

rodney528 (approved):

PR #13304, opened 2023-09-19
Implement hscript-improved

TheLeerName:
as u can see in this [repo](https://github.com/TheWorldMachinima/SScript), SScript is not exists anymore, so i got idea implement `hscript-improved` lib from Codename Engine cuz it have the same things as in SScript like `import` and such

So **now** with this library we can use this script:
```hx
function onCreatePost() {
	gf.visible = false;
}
```
and yes, as you can see, we can avoid using `game` variable 😱
classes still bugged and dont work normally (skill issue)
also i rewritted error messages (lol as in #13017), and i know about bug with duplicated file name in haxe 4.3.X lol
also `getTextFromFile` lua function and `Paths.getTextFromFile` now has 3rd argument, if it `true`, then it gets file by absolute path
also `addHaxeLibrary` can import enums now

### UPD: 17:49 - 19.09.2023
- errors on hscript now traces to console
- if u got error in `runHaxeCode`, it stops executing
- `Lua_helper.add_callback()` now just `set()`
- `setOnLuas` now not crashes when u trying to set function with it
- trace of elapsed time after loading script

## WARNING: USE THIS COMMAND BEFORE COMPILING PLS
```
haxelib git hscript-improved https://github.com/TheLeerName/hscript-improved
```
*will use original hscript-improved if my pr in original repo will be merged*

PR #13325, opened 2023-09-23
rewrite error messages in scripts

TheLeerName:
also init of lua/hx scripts now traces elapsed time on loading

no stupid cuts when printing for example unknown variable name

originally from #13304


Karim-Akra (approved):
nice

TheLeerName:
not interested

PR #13326, opened 2023-09-23
lua fixes & changes

TheLeerName:
- u can use `set()` instead of `Lua_helper.add_callback()` in `psychlua.FunkinLua`
- `setOnLuas()` lua / playstate function now not crashes on function set
- created `import.hx` in `psychlua` package cuz we need `llua` package only in `psychlua`
- `backend.Discord.DiscordClient.addLuaCallbacks()` now uses `psychlua.FunkinLua` as first argument instead of `llua.State`

originally from #13304


JohnDoe3453:
i think its useful

lunacynne (approved):

TheLeerName:
not interested

PR #13367, opened 2023-09-30
let me use decimal bpms in titlestate!!!!!

libauror-a:
changes TitleData's bpm field to a float like every other instance of bpm in the engine


moxie-coder (approved):
agreed

PR #13384, opened 2023-10-04
Auto sustain scale adjustment using frameHeight

MaybeMaru:


DetectiveBaldi (approved):
the simpler the math the better 😁

PR #13250, opened 2023-09-06
Add softcoded achievements and update SScript to 5.2.0

ghost:
Added softcoded achievements, supports both SScript and PsychLua, added example in modTemplate zip.
Currently undocumented

PR #13409, opened 2023-10-08
Custom achievements for typedef'ed achievement system

ghost:
Supports both SScript and PsychLua. Added example in modTemplate zip.
Only difference with [this](https://github.com/ShadowMario/FNF-PsychEngine/pull/13250) PR is that `index` variable is removed due to new achievement system not supporting indexing.


crowplexus:
I'd say wait for the next experiemntal commit since the achievements system changed a little bit since then

TahirKarabekiroglu:
> I'd say wait for the next experiemntal commit since the achievements system changed a little bit since then

I am gonna update it when the next commit comes out.

charlesisfeline (approved):
i hope it doesnt get deleted this time i mean you switched accounts so much and i can see why

nice pr tho

PR #13433, opened 2023-10-12
Camera fix for Flixel's new version

NexIsDumb:
Yeah they changed the camera scroll method: https://github.com/HaxeFlixel/flixel/pull/2920
So now without this fix the camera is slow as fuck when on high fps since maths has changed
I also needed to set the fixed timestamp as true otherwhise sometimes the camera would have just randomly teleported

### DIFF
--- a/source/states/AchievementsMenuState.hx
+++ b/source/states/AchievementsMenuState.hx
@@ -203,7 +203,7 @@ class AchievementsMenuState extends MusicBeatState
 			}
 		}
 
-		FlxG.camera.followLerp = FlxMath.bound(elapsed * 9 / (FlxG.updateFramerate / 60), 0, 1);
+		FlxG.camera.followLerp = FlxMath.bound(elapsed * 9 * (FlxG.updateFramerate / 60), 0, 1);
 
 		if (controls.BACK) {
 			FlxG.sound.play(Paths.sound('cancelMenu'));
--- a/source/states/MainMenuState.hx
+++ b/source/states/MainMenuState.hx
@@ -137,7 +137,7 @@ class MainMenuState extends MusicBeatState
 			FlxG.sound.music.volume += 0.5 * elapsed;
 			if(FreeplayState.vocals != null) FreeplayState.vocals.volume += 0.5 * elapsed;
 		}
-		FlxG.camera.followLerp = FlxMath.bound(elapsed * 9 / (FlxG.updateFramerate / 60), 0, 1);
+		FlxG.camera.followLerp = FlxMath.bound(elapsed * 9 * (FlxG.updateFramerate / 60), 0, 1);
 
 		if (!selectedSomethin)
 		{
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -1594,7 +1594,7 @@ class PlayState extends MusicBeatState
 
 		FlxG.camera.followLerp = 0;
 		if(!inCutscene && !paused) {
-			FlxG.camera.followLerp = FlxMath.bound(elapsed * 2.4 * cameraSpeed * playbackRate / (FlxG.updateFramerate / 60), 0, 1);
+			FlxG.camera.followLerp = FlxMath.bound(elapsed * 2.4 * cameraSpeed * playbackRate * (FlxG.updateFramerate / 60), 0, 1);
 			if(!startingSong && !endingSong && boyfriend.animation.curAnim != null && boyfriend.animation.curAnim.name.startsWith('idle')) {
 				boyfriendIdleTime += elapsed;
 				if(boyfriendIdleTime >= 0.15) { // Kind of a mercy thing for making the achievement easier to get as it's apparently frustrating to some playerss
--- a/source/states/TitleState.hx
+++ b/source/states/TitleState.hx
@@ -83,7 +83,6 @@ class TitleState extends MusicBeatState
 		#end
 		Mods.loadTopMod();
 
-		FlxG.fixedTimestep = false;
 		FlxG.game.focusLostFramerate = 60;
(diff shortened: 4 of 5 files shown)


Raltyro:
Oh nice they fixed the camera

Raltyro:
@Rudyrue ????? RUDY

Raltyro (approved):

Rudyrue:
me

Rudyrue:
also yea they did

that's only because of me tho and my old fix is kinda better ?? doesn't account for lag spikes tho

NexIsDumb:
> also yea they did
> 
> that's only because of me tho and my old fix is kinda better ?? doesn't account for lag spikes tho

well i have a video, but its not lag dont get me wrong
if you try the code i fixed but then put thr fixed timestep as false sometimes it would just like teleport
it didnt always happen, but when it did it was noticeable and annoying

NexIsDumb:
oh wait shit you were talking about your fix
LMAO SORRY i didnt read well with the sun on the screen

KarimAkra (approved):

Vortex2Oblivion (approved):

BlueColorsin (approved):
I HATE the slow camera 10/10

NexIsDumb:
> I HATE the slow camera 10/10

real

Geokureli:
oh man, I remember being iffy on this change, but ultimately thinking that it would only affect niche cases. let me know in the future when i break yo stuff

NexIsDumb:
> oh man, I remember being iffy on this change, but ultimately thinking that it would only affect niche cases. let me know in the future when i break yo stuff

👍

nubz4lif (approved):

Geokureli:
Heads up, i have [this camera change](https://github.com/HaxeFlixel/flixel/pull/2922) slated for flixel 6.0.0. I've labelled it a breaking change, it wont break compilation or cause runtime errors but if your game has `FlxG.fixedTimestep = false;` (`true` is the default), you camera may lerp faster or slower depending on performance

Raltyro:
the fixedTimeStep is really needed for psych :(
Without it, most of lua script and music sync wont work

NexIsDumb:
> Heads up, i have [this camera change](https://github.com/HaxeFlixel/flixel/pull/2922) slated for flixel 6.0.0. I've labelled it a breaking change, it wont break compilation or cause runtime errors but if your game has `FlxG.fixedTimestep = false;` (`true` is the default), you camera may lerp faster or slower depending on performance

> the disabled fixedTimeStep is really needed for psych :(
Without it, most of lua script and music sync wont work

Yup, in fact I've enabled it in this PR
Even before you made that change for 6.0.0 if the timestep was disabled, the cmaera would have done odd stuff like speed increasing and decreasing randomly, causing it to teleport lmfao

NexIsDumb:
On line 86 of the Titlestate.hx

Geokureli:
> > Heads up, i have [this camera change](https://github.com/HaxeFlixel/flixel/pull/2922) slated for flixel 6.0.0. I've labelled it a breaking change, it wont break compilation or cause runtime errors but if your game has `FlxG.fixedTimestep = false;` (`true` is the default), you camera may lerp faster or slower depending on performance
> 
> > the disabled fixedTimeStep is really needed for psych :(
> > Without it, most of lua script and music sync wont work

I can do some more testing here to see how this behaves, and it would actually be very helpful if someone tried [that branch](https://github.com/HaxeFlixel/flixel/pull/2922) on their project with some induced performance lag. I really do think this change will result in a smoother camera, but currently, thats a theory.

To summarize: while the change makes the lerp change based on current performance, it also makes different fps have less of a drastic effect on lerps value. For example,  previously, a lerp of 0.5 or higher would effectively snap 100% to the target position every frame at 30fps or les, now a lerp of 0.5 at 30fps will move 75% to the target every frame, which more closely matches the behavior you'd see at 60 fps where it would move 50% twice (half plus half of a half which 75%)

> Yup, in fact I've enabled it in this PR Even before you made that change for 6.0.0 if the timestep was disabled, the camera would have done odd stuff like speed increasing and decreasing randomly, causing it to teleport lmfao

That's really odd, considering that both the old and new code don't depend on the elapsed time, they use `followLerp` and `FlxG.updateFramerate`, so it should apply the same lerp ratio unless either of these are changed per-frame. I suspect that the game may be eating or skipping render frames when performance is bad enough, and that's what's causing the behavior you describe

NexIsDumb:
> That's really odd, considering that both Flixel's old and currently released lerp code don't depend on the elapsed time, they use followLerp and FlxG.updateFramerate, so it should apply the same lerp ratio unless either of these are changed per-frame. I suspect that the game may be eating or skipping render frames when performance is bad enough, and that's what's causing the behavior you describe

It may be true cause it was mainly happening very few times and thats when @NeeEoo told me to try enabling fixedtimestep (It was happening with every FPS option, from 60 up to 240) and my PC its actually a bit potato.
After enabling fixedtimestep ofc those "teleportation" problems disappeared

Geokureli:
I'm always using latest everything, to my knowledge

NexIsDumb:
> I'm always using latest everything, to my knowledge. I may need to update that branch with changes to dev **Edit:** updated

Alright, tested
I'm gonna post the results in the your PR since it's more related

ShadowMario:
Had to extend FlxCamera for `backend.PsychCamera` so I can make the camera use delta for properly working on higher framerates without randomly snapping at 240FPS and shit, this works better than your PR.
Done at commit [#799630a](https://github.com/ShadowMario/FNF-PsychEngine/tree/799630a10acad1a221edff5da552ffa22bc88fdc)

PR #13469, opened 2023-10-19
License Change

xiaochenhao576:
I found that the license just copy and forgot to **change something**(it was the [yyyy] [name of copyright owner]),so i did it lol,also,it was created at 2y ago,so [yyyy] will be 2021

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2021 ShadowMario
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


NexIsDumb:
i dunno if itd be wise to edit it, but i think at thia point its better to copyright ninjamuffin for the original fnf idk

EyeDaleHim:
i don't think you should do this, i think it's better that shadow mario does this himself instead of a random PR.

NexIsDumb:
> i don't think you should do this, i think it's better that shadow mario does this himself instead of a random PR.

indeed

ShadowMario:
I do not own Friday Night Funkin', I think it's more reasonable to credit the Funkin' Team, however I understand absolutely zero things of copyright laws, so I will just leave it blank as I don't care enough.

PR #13504, opened 2023-10-27
fix path for mobile

HomuHomu833:
I think ShadowMario will deny this due mobile compilability but this not means everything in mobile is there. It's just path fix for easy porting ig idk


KarimAkra (approved):

ShadowMario:
surprise, bucko

mcagabe19no2fa:
insane.

PR #13260, opened 2023-09-09
more platfroms in getBuildTarget

ghost:
probably this is will be good idk

PR #13530, opened 2023-10-31
some platfroms for getBuildTarget

ghost:
old pr:https://github.com/ShadowMario/FNF-PsychEngine/pull/13260

This happend due I accidently removed the branch (and can't restore blame github)


ShadowMario:
This variable is kind of just a generalization thing, so it doesnt need to be *that* specific

PR #13532, opened 2023-11-01
better hitsound

KarimAkra:
yes i grabbed this from [ChartingState](https://github.com/ShadowMario/FNF-PsychEngine/blob/main/source/states/editors/ChartingState.hx#L2115)


ghost (approved):

Karim-AkraNo2FA:
Found out that this doesn't let you have a specific volume for the hitsound so I'll close it for now

PR #13534, opened 2023-11-01
Fix Memory Counter in FPS Overlay showing greater memory usage than intended

crowplexus:
this happened due to the way `memoryMegas` was being calculated, of which made the display show slightly more memory than usual.

Left is new usage | Right is old (pre-fix) usage
image

this was brought to my attention by @NeeEoo and its very much the reason why I decided to open the PR

this change also makes it so the overlay only updates every second instead of frame, since there'd be a slight overhead in performance by updating it every frame, and who needs it anyways? :p

this also *technically* allows for scripts to make a `memoryPeak` variable now since I made `memoryMegas` public

### DIFF
--- a/source/openfl/display/FPS.hx
+++ b/source/openfl/display/FPS.hx
@@ -111,6 +111,6 @@ class FPS extends TextField
 		}
 
 		cacheCount = currentCount;
-		deltaTimeout += delta;
+		deltaTimeout += deltaTime;
 	}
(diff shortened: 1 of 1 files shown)


NexIsDumb (approved):

TheLeerName (approved):
ye

ghost (approved):
I agree

PR #13541, opened 2023-11-02
Fix of charting state doing Paths.formatToSongPath() two times

TheLeerName:
for example we have: `M.I.L.K` song, but when game tries load it in playstate it loads from folder `mi.l.k`, but in charting state it loads from `mil.k`, cuz it doing Paths.formatToSongPath() of song name two times

this pr fixes that

btw you can just fix it by adding `g` to vars in `Paths.formatToSongPath()` function but it will break parsing some songs
GitHubDesktop_84KfroYHP7


TheLeerName:
bro pls import flxanimate

PR #13585, opened 2023-11-09
Set pixel antialiasing and scaling rules for `stageUI` if it ends with `-pixel`

crowplexus:
I should've done this earlier lol


crowplexus:
reopened cuz single commit feels nicer lol

PR #13586, opened 2023-11-09
Make the Score Update function a dynamic so anyone can override it

crowplexus:
people kept blabbering on how it was broken in lua so here u go

override it yourself now.


crowplexus:
last one is because I'm stupid and forgot to make a group for the notes so they wouldn't render on top of the HUD

charlesisfeline (approved):
real shi-

its good!!!!!!!!!!!!!

PR #13634, opened 2023-11-18
Update FlxSound.hx

Jamextreme140:
OpenFL 3.9.2 -> SoundChannel class overhauled: "__source" private variable renamed to "__audioSource"


richTrash21:
fixed in flixel 5.0.0

PR #13666, opened 2023-11-24
Fixes the broken Syntax in PlayState

KarimAkra:
all that time and we were blaming VSCode for this


ghost (approved):

charlesisfeline (approved):

CaptainBaldi:
this isnt really broken syntax
its bad formatting
also there's bad formatting in psych everywhere this does like nothing

PR #13674, opened 2023-11-25
make icons in freeplay centered properly

Rudyrue:
icons in freeplay are off center
this fixes that

# without fix

# with fix

icons are also both 300x150

this might also be a shit way of fixing it but im sure someone will make a better method down the road lmao


Unholywanderer:
what if instead i snapped you into 2 pieces like a kit kat bar

Rudyrue:
nerd emoji's you

richTrash21:
mid😔

Rudyrue:
ok bro

PR #13679, opened 2023-11-26
Add helpful trace information to `Paths.image`

DetectiveBaldi:
`Paths.image` currently does not output much useful information when an image you are trying to load is missing. It only tells you what image couldn't be found, and no other information.

This PR adds a new parameter to Paths.image, `posInfos`. This parameter is not meant to be filled in and is recommended to be left empty as much as possible. However, you SHOULD be able to insert your own data without much trouble.

The trace has been extended to display the class name you are calling the function from, as well as the line number, here's a quick sample.
image

### DIFF
--- a/source/backend/Paths.hx
+++ b/source/backend/Paths.hx
@@ -253,7 +253,8 @@ class Paths
 			if(retVal != null) return retVal;
 		}
 
-		trace('oh no its returning null NOOOO ($file) (' + posInfos.fileName + ', ' + posInfos.lineNumber + ')');
+		trace('Image with key "$key" could not be found' + (library == null ? '' : ' in the library "$library"') + '! ' + '(${posInfos.fileName}, ${posInfos.lineNumber})');
+		
 		return null;
 	}
 
--- a/source/backend/Paths.hx
+++ b/source/backend/Paths.hx
@@ -220,7 +220,7 @@ class Paths
 	}
 
 	public static var currentTrackedAssets:Map<String, FlxGraphic> = [];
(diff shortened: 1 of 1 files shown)


libauror-a (approved):
https://github.com/ShadowMario/FNF-PsychEngine/assets/75984453/225658f1-9a30-4602-8394-e9a8884da17b

source/backend/Paths.hx
```diff
@@ -253,7 +253,7 @@ class Paths
 			if(retVal != null) return retVal;
 		}
 
-		trace('oh no its returning null NOOOO ($file)');
+		trace('oh no its returning null NOOOO ($file) (' + posInfos.fileName + ', ' + posInfos.lineNumber + ')');
```
Geokureli, line 256: instead try 
```suggestion
		FlxG.log.warn('Could not find image with key: "$key"' + (library == null ? "" : 'in library: "$library"'));
```

also you can have cleaner string interpolation with ${}
example:
```hx
trace('image not found ($file) (${posInfos.fileName}, ${posInfos.lineNumber})');
```
though i don't think, posInfos should be output in FlxG.log.warn

Geokureli: i want to be clear, I am not a maintainer of this repo, just giving some feedback

ender69funk:
idk but it's a useful or not

Geokureli:
I still say trace should be avoided at all cost, especially for errors, either throw an error or use FlxG.log.warn/error if the game can continue

PR #13572, opened 2023-11-06
More finalizations (Flxanimate & new setup method)

NexIsDumb:
Both finalizes #13543 and #13257

Why am I deleting the installer once it finishes downloading? Speaks for itself lmfao; it's literally useless keep it there once the installation was completed

PR #13681, opened 2023-11-26
Skip SScript setup mode in action builds

nubz4lif:
This will skip the setup mode introduced in SScript 7.7.0

Note that the action build will still fail due to https://github.com/ShadowMario/FNF-PsychEngine/pull/13572 not being merged

Theoretically this should be the equivalent of saying no to customizing settings in setup mode


ghost (approved):
most useless macro™

PR #13684, opened 2023-11-26
rewrite most of MainMenuState

Rudyrue:
makes some variables final
sets active to false on the bg and text objects cuz they don't need to be updated every frame
reworks how curselected is handled which now makes adding mouse support easier

also fixed the cam lerp bug that i caused because psych does cam lerp weirdly

also also yea i could've used setFormat for the text shur up.......


DetectiveBaldi (approved):
this has reminded me that i should use finals more in the main menu im making

richTrash21 (approved):
FlxMath.wrap()????? so based
ezgif-2-db4f8e0187

Rudyrue:
oh my GOD i hate github so fucking much

Rudyrue:
welp that's stuck in this pr now

PR #13686, opened 2023-11-27
Better health system.

ghost:
Reduces health value checks down to a single line, using FlxMath.bound, also accounts for practice mode when the health hits/is at 0.

### DIFF
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -1632,12 +1632,7 @@ class PlayState extends MusicBeatState
 		iconP2.updateHitbox();
 
 		var iconOffset:Int = 26;
-		if (healthBar.bounds.max != null) {
-			if (health > healthBar.bounds.max) health = healthBar.bounds.max;
-		} else {
-			// Old system for safety?? idk
-			if (health > 2) health = 2;
-		}
+		health = FlxMath.bound(health, 0, (healthBar.bounds.max != null ? healthBar.bounds.max : 2));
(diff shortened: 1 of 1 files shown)


richTrash21 (changes requested):
I suggest using Math.min(), since health <= 0 is always gameover.
Ex:
```haxe
health = Math.min(health, (healthBar.bounds.max != null ? healthBar.bounds.max : 2));
```
Also it will be a good idea to add a setter for the health and move bound code there, so that hp would be bounded only when it's value changed.
Ex:
```haxe
public var health(default, set):Float = 1.0;

function set_health(HP:Float):Float
{
    if (health == HP)
        return HP; // hp hasn't changed = just return it

    // bounds HP to the lower bound
    health = Math.min(HP, (healthBar.bounds.max != null ? healthBar.bounds.max : 2));
    doDeathCheck(); // just a convinience
    return health; // returns bounded health
}
```

Blue2359:
> `I suggest using Math.min(), since health <= 0 is always gameover. Ex:`
My fix also accounted for practice mode, since you can't die when it's enabled, and the icons just start going offscreen when the `health` is below 0.

richTrash21:
> My fix also accounted for practice mode, since you can't die when it's enabled, and the icons just start going offscreen when the `health` is below 0.

Ohhh yeah, forgot about this. My bad

ShadowMario:
Health is supposed to get negative

PR #13691, opened 2023-11-28
Automatically compile and upload builds on release

nubz4lif:
Automatically runs build.yml on release and uploads the artifacts to the release

For this to work, build.yml needs to compile everything successfully, which isn't true right now because:
- https://github.com/ShadowMario/FNF-PsychEngine/pull/13572 isn't merged
- Library versions are not locked in hmm.json
- Windows builds fail due to missing discord_rpc for some reason even if the other 2 issues are fixed (linux and mac compiles successfully though)

[Here is what it looks like if it works successfully using my repo](https://github.com/Noobz4Life/FNF-QuackEngine/releases/tag/0.2)

(also I understand this will be useless due to ShadowMario stopping development soon, but it might be useful for forks)


Noobz4Life:
also i'll try to update this for 32-bit builds at some point, i just don't find it a priority right now since barely anyone uses 32-bit operating systems anymore

Noobz4Life:
moved to https://github.com/ShadowMario/FNF-PsychEngine/pull/13940

PR #13718, opened 2023-12-06
Why is playbackRate put twice?

rodney528:


ShadowMario:
It's in purpose, because when i tested it (might've been an issue with older flixel or something?) it didn't look like it was on the proper speed

RodneyAnImaginativePerson:
> It's in purpose, because when i tested it (might've been an issue with older flixel or something?) it didn't look like it was on the proper speed

Huh... wierd.

PR #13721, opened 2023-12-06
removed " from hitsound volume

ender69flock:
nice one


NexIsDumb:
correct me if im wrong but at this point shouldnt you also correct does with do? cuz its plural

ender69funk:
idk

PR #13626, opened 2023-11-17
Music Player V2

ghost:
Music Player V2 fixes tons of bugs and adds new features while staying minimalistic and easy to use.

New features include:
- Pitch modifier (Maximum 3x, Minimum 0.05x)
PsychEngine 2023-11-18 02-35-58-846
- Progress Bar
PsychEngine 2023-11-18 02-36-13-066
- Reset feature which resets song position and playback rate (pitch)

https://github.com/ShadowMario/FNF-PsychEngine/assets/146023700/0b644bda-e4c5-4e48-99ef-4bcd7a82f2ee

- Vocals resync if they are out of sync with the instrumental
- Actual support for songs without vocals
- Auto pause disabling when playing a song
- If returned to main menu without a song playing, main menu music will play back again

PR #13723, opened 2023-12-06
Music Player V2.5

ThomasDarkson:
This PR is almost the same as #13626 except it fixes even more bugs and optimizes a little more
also updates tahir credit


NexIsDumb (approved):
looks good to me

PR #13748, opened 2023-12-10
Fix FPS Counter using the wrong value for clamp

crowplexus:
yea


ghost (commented):
that's why fps always goes up and stucks at ur selected framerate

ghost (approved):

PR #13756, opened 2023-12-13
update bug report template

crowplexus:
this is to include a more in-depth view as to how the users are using the engine and how they stumbled upon the bug


CobaltBar:
**All checks have failed**

crowplexus:
> **All checks have failed**

they won't work because this is being commited to main

PR #13759, opened 2023-12-13
"no-response" v2 - Close issues after 10 days of inactivity

crowplexus:
[actions/stale](https://github.com/actions/stale) is a bot that helps in closing isues after being marked as `stale`

an issue gets marked as so when it becomes inactive for 5 days, this can be changed in the file

after being marked as stale, it waits 5 more days, then closes the issue if the label remains there

this workflow will ignore issues labelled as "enhancement", "need more information" and "special"


crowplexus:
damn thanks github (reopening shortly)

PR #13777, opened 2023-12-17
Optimize MainMenuState

CCobaltDev:
Removed the pointless loops, and removed centering the **sprites every update**


AbdooOwd:
I never thought of that. That's rlly cool dude!
Like fr

PR #13781, opened 2023-12-18
hscript fixes

Rudyrue:
fixes stuff like `add()` and `remove()` not returning the said object like in `FlxGroup`

completely fixes hscript error handling, making it print to the console 100% of the time and also making the errors less dependent on lua at the same time


Rudyrue:
idk if this breaks runHaxeCode in lua or not but i didn't test it cuz i don't really care that much for runHaxeCode since i mainly just use pure hscript now

Unholywanderer:
nerd

Rudyrue:
nuhuh

PR #13795, opened 2023-12-20
editor dropdownmenu fix

FixedData:
fixes drop down menu not working correctly at different zooms


FixedData:
also fixes the the slider breaking at different zooms

PR #13825, opened 2023-12-23
Replace Every X++ With ++X

KarimAkra:
heared that ++X is faster a bit compared to X++ so i did this


ghost (approved):

cyn0x8:
There isn't actually a difference in performance here... Even if there was, be careful when replacing them since they return different things; `x++` gives `x` before the increment, and `++x` gives `x` after the increment.

ShadowMario:
Technically speaking its nearly the same thing, the difference in performance would be less than a microsecond too

PR #13831, opened 2023-12-23
Install `hxdiscord_rpc` instead of using git and cleaned `project.xml` a bit.

MAJigsaw77:


mcagabe19:
@MAJigsaw77 Linux builds dying due `linc_luajit` thay's why we add `NO_PRECOMPILED_HEADERS` (same thing we did before in android gcc ndk's)

MAJigsaw77:
> @MAJigsaw77 Linux builds dying due `linc_luajit` thay's why we add `NO_PRECOMPILED_HEADERS` (same thing we did before in android gcc ndk's)

I understood but it kills performance...

PR #13840, opened 2023-12-24
Gameover Improvements

Trayfellow:
- Fix fadeIn for Week 7 so it doesn't tween abruptly
- Fix deathConfirm offsets & Camera position being off-centered for gameover Pixel BF
- Positioning for the gameover sprites should almost be the exact same as their alive counterparts now
- Timer length during fade out is now based on how long your end sound is (gameOverEnd files were modified to cut off any empty parts in the sound)
- Code cleanup so the animation finish check uses finishCallback instead


CaptainBaldi:
You might want to change the branch to main but this seems pretty good

KarimAkra:
hope this fixes the memory leak i was having when opening GameOverSubState with runHaxeCode lmao

ACoolioDude (approved):
This is not bad. It will help for memory peak

NexIsDumb:
> This is not bad. It will help for memory peak

huh, what does the memory peak have to do here

Stefan2008Git:
> > This is not bad. It will help for memory peak
> 
> huh, what does the memory peak have to do here

I mean, the memory will be the less big i want to say. Sorry

NexIsDumb:
> > > This is not bad. It will help for memory peak
> > 
> > 
> > huh, what does the memory peak have to do here
> 
> I mean, the memory will be the less big i want to say. Sorry

nono, i got that, but why? 2 files have been reduce a little, while others 2 the size has been "drastically" increased

PR #13843, opened 2023-12-24
Mod Settings in Pause Menu (FAIL PR 2)

Ramiro0045:
it adds a mod settings to the pause menu
and makes the enable/disable buttons dont appear if you are on playstate to probably prevent a crash

Preview: [https://youtu.be/GeIlMPp8IWA](url)


Ramiro0045:
ahh, i think i messed up on something, i dont have time to like fix it ahh

Ramiro0045:
Made a bit of a mistake on this, ill try to fix it later in the proper branch and hopefully it gets approved
will be kinda neat being like in the credits even tho this is basically nothing compared to the other people that are in the credits

PR #13851, opened 2023-12-25
Mod Settings in Pause Menu 0.7.2

Ramiro0045:
it adds a mod settings to the pause menu
and makes the enable/disable buttons dont appear if you are on playstate to probably prevent a crash

Preview: https://youtu.be/GeIlMPp8IWA]


rodney528 (approved):

Ramiro0045:
for some reason, the video thing wasnt working, but i fixed it

KarimAkra:
wouldn't it fit better making the "options" option open a new section like the difficulty thing with 2 options one switchs the state to OptionsState and one just opens the mod's settings substate directly instead of taking you to ModsMenuState

Ramiro0045:
I could maybe do that but i didnt see it as neccesary
also, clicking options just to click options again will be a extra click that is not neccesary
so is easier to do it in this way (also easier to code it this way) than grouping both options into one button

also, it will probably mess with some lua scripts that make custom menus
(idk if they are able to delete or add buttons in just lua and/or rch)

i know someone else in the discord told me about adding the mod settings in the options state but again i see it as unnecesary
both things are options but those two change different stuff and even tho i could group them, it will be a extra click
and in difficulties, it does make sense bcs who knows how many difficulties and what difficulties will be in the song
while it will always be two types of options

PR #13853, opened 2023-12-25
Update hmm.json to use the version data from the wiki

EliteMasterEric:
`hmm.json` now pins all mods to specific library versions (for Haxelib libraries) and commit hashes (for Git libraries). This ensures that any user using `haxelib run hmm install` to install libraries is receiving the exact same library versions, and library updates will not cause errors for newer users.

Steps should definitely be taken to transition `setup` away from a list of `haxelib install` commands and towards a setup which simply installs and executes `hmm`.


EliteMasterEric:
Also what is the motivation behind not using `hmm` for Github Actions? @crowplexus since git blame says you made `setup/Main.hx` which literally just does what hmm is supposed to do but you rewrote it yourself for some reason.

crowplexus:
> Also what is the motivation behind not using `hmm` for Github Actions? @crowplexus since git blame says you made `setup/Main.hx` which literally just does what hmm is supposed to do but you rewrote it yourself for some reason.

github actions, it might've been easuer to execute hmm tbh, change was sort of pointless lol

feel free to take it off tho

EliteMasterEric:
While testing elsewhere I discovered the issue which prompted the switch away from `hmm`.

`hmm` utilizes the project-local Haxe repository feature of core Haxelib. This feature appears to have an issue where it does not function properly when attempting to include libraries which make use of native code (such as `hxdiscord_rpc` and `hxCodec`).

As such, I am marking this PR as a draft until such a time that the issue can be fixed.

Definitely feel free to copy the `hmm.json` file over though.

EliteMasterEric:
Watch for resolution of this issue: HaxeFoundation/haxelib#618

crowplexus:
still waiting for a proper resolution to this issue, I'd come in handy right now 😅

PR #13865, opened 2023-12-27
Fix FPS being incorrectly calculated

crowplexus:
hi rudy hi zyfix

### DIFF
--- a/source/debug/FPSCounter.hx
+++ b/source/debug/FPSCounter.hx
@@ -56,7 +56,7 @@ class FPSCounter extends TextField
 		times.push(now);
 		while (times[0] < now - 1000) times.shift();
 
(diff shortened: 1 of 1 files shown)


Rudyrue:
imp emoji

Unholywanderer:
crow!!! you gotta add something like 
`currentFPS = times.length < FlxG.updateFramerate ? times.length : FlxG.updateFramerate;`
or it still wont update because currentFPS wont ever be lower than update framerate after its initially set and reaches it i think idk
it just didnt update until i did this 😔

mcagabe19:
> imp emoji

👿

Stefan2008Git:
I imported to my modified Psych 0.7.1h build, but the FPS is staying on 60 framerates for some reason @crowplexus. Do you know why is happening that?

PR #13882, opened 2023-12-29
Adds VideoManagers classes and littel support for VideoSprites

KarimAkra:
meant to manage different hxCodec versions while keeping the same usage functions

ye i did pull request this once before but my github got flagged so i assume the pull request got removed by github..

i did rework alot of things compared to the old pr so it's not just a copy-paste of the old thing

i'm not sure if VideoSpriteManager should stay but i'm ready to remove it if i was asked to..

only issue with this is that for some reason the game will freeze if the video that's supposed to be played is missing, i tried alot to fix it but nothin' worked so i gave up on that

### DIFF
--- a/source/backend/VideoManager.hx
+++ b/source/backend/VideoManager.hx
@@ -1,0 +1,80 @@
+package backend;
+
+#if VIDEOS_ALLOWED 
+#if (hxCodec >= "3.0.0") import hxcodec.flixel.FlxVideo as VideoHandler;
+#elseif (hxCodec >= "2.6.1") import hxcodec.VideoHandler;
+#elseif (hxCodec == "2.6.0") import VideoHandler;
+#else import vlc.MP4Handler as VideoHandler; #end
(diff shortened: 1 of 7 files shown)


ghost (approved):

CaptainBaldi:
are there Lua functions for VideoSprites here? Otherwise, I don't see the point of all these managers.
All of this could be shortened down to a map containing video's/videosprites in PlayState, which would make videos much more accessible to Lua (pausing, resuming, changing pitch) etc
Then, import VideoSprite in PlayState or elsewhere so it can be accessed via hscript, and then theres pretty much no point for the managers

KarimAkra:
i honestly didn't think about making them directly usable in Lua
might do this later...

KarimAkra:
i'll be adding some lua functions for video sprites so this will be a draft for a while..

roma-perec-bp (approved):

PR #13883, opened 2023-12-29
fixes global scripts running twice

KarimAkra:


ghost (approved):
karim said to me approve blame him-

KarimAkra:
> karim said to me approve blame him-

ShadowMario:
If I am correct, it only would run global scripts twice if said scripts are from a week inside the global mod's folder, right?

KarimAkra:
ye they run twice in the global mod only when they're in the mod's scripts folder

PR #13885, opened 2023-12-29
fixes crashing when week json has invalid fields

KarimAkra:


ShadowMario:
I really need to know, in what circumstances would you have invalid fields? Because if you use the week editor it should never give this sort of error

KarimAkra:
just though it would be safer having this check
i personally was having this as an issue because of a mod that was made in a modified psych engine build that has extra week variables
using the mod in the original builds was fine until WeekData started using Reflect for getting WeekFile variables it would result in a crash

PR #13905, opened 2024-01-03
Remove Asset Libraries

SrtHero278:
A small quality of life commit for source assets.
`week_assets` is merged with the default library. and the shared library is no longer required.
All moved assets have little to no change to softcoded modding and has been tested in all of the story mode weeks.


crowplexus (approved):
seems nice and doesn't seem like it'd break mods

PR #13913, opened 2024-01-04
Fix CoolUtil.openFolder using the wrong variable for Linux

Trayfellow:
Why was it like that


ghost (approved):
💀

charlesisfeline (approved):
i think the function was already unused but ok

PR #13949, opened 2024-01-11
Revert "Load pause music asynchronously"

superpowers04:
This implementation is a bit off. If the player somehow closes the pause menu before the song is loaded, it'll play the song during the game. From my understanding the song will only have to load once and then it'll be pre-cached, so this is kind've a problem that doesn't need fixing imo

### DIFF
--- a/source/substates/PauseSubState.hx
+++ b/source/substates/PauseSubState.hx
@@ -12,11 +12,6 @@ import states.StoryMenuState;
 import states.FreeplayState;
 import options.OptionsState;
 
-#if (target.threaded)
-import sys.thread.Thread;
-import sys.thread.Mutex;
-#end
-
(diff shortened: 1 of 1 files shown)


Rudyrue:
not even that
the haxe threading class is extremely buggy to my knowledge and doesn't work properly, and sometimes just straight up crashes the game

it'd be better to use [lime.app.Future](https://github.com/openfl/lime/blob/master/src/lime/app/Future.hx)
but even when using that

it's basically useless
the only reason i can think of using threads for this is because when opening the pause menu the game freezes for like 2 seconds because it's loading the music file

but it doesn't even freeze
because it caches the music file on PlayState create()
so there's no reasonable reason to have a seperate thread for this

PR #13958, opened 2024-01-13
Fix PsychLua getRandomType functions never being able to return 0

Rudyrue:
ok so short answer
psych random functions never return 0
so i fixed it

long answer
since the `exclude` variable is usually null (no one ever sets it anyways) the number that gets pushed to the exlcude array is always 0
because it runsa for loop based on the length of the exclude array
which is 0
because it's null

and because the only number that's pushed to the exclude array is 0 the random functions aren't able to ever return 0


melodiekit (approved):
cool

PR #13995, opened 2024-01-18
move `zoomTxt` to `bpmTxt`

system32unknown:
because the fps counter covers below `zoomTxt`, they difficult to see zoom text.


BombasticTomDev (approved):
yessir

ghost (approved):

PR #14008, opened 2024-01-19
Fixes some issues caused by split vocals

KarimAkra:
playing without split vocals was causing the game to resync vocals almost every frame for some reason so when your FPS drops a bit it could cause some sort of sound stuttering with the vocals, so this here fixes it
this also fixes player vocals volume still being 0 even after the opponent hit a note(while playing without split vocals ofc)

### DIFF
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -1555,12 +1555,7 @@ class PlayState extends MusicBeatState
 		if (paused)
 		{
 			if (FlxG.sound.music != null && !startingSong)
-			{
-				var vocalsToResync:Array<FlxSound> = [vocals];
-				if(splitVocals)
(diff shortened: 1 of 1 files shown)


Trayfellow (approved):
nice, fixed the bug i had with week 6's vocals being stuttery after i paused and then resumed the song

ghost (approved):

cabfile:
does this fix the bug where the voices glitch if they're shorter than inst?

mcagabe19:
no

cabfile:
is that ever going to be fixed

ShadowMario:
Does not fix the stuttering issue, as vocals are still being constantly resynced, so I'm closing this PR, feel free to make a new one if you find a definitive fix.

PR #14056, opened 2024-01-28
take scale factors into account when offsetting characters

crowplexus:
fixes having to reoffset them at all when changing scale

demo:

https://github.com/ShadowMario/FNF-PsychEngine/assets/45212377/a1df3329-eb73-460d-9483-c5d3b4ba76c3


ghost (approved):

NexIsDumb (approved):
oh
ive noticed this few mins ago
i was about to make a fix too LOL

lunacynne (approved):

crowplexus:
unfortunately would break a lot of older mods

tposejank:
> unfortunately would break a lot of older mods

could've updated it only in the character editor explicitly when changing scale of something instead of everywhere
instead of using the scale everywhere, precalculating the value in the character editor

PR #14064, opened 2024-01-29
Fixing the layers because they are broken

ACoolioDude:
It's seems like Shadow Mario did something really weird to make note only to show outside from layer instead to show ui group outside. You know y'all, notes and splashes are should be inside on layers and ui group should be outside


Stefan2008Git:
If you hate this pull request, i will close it. It's not a problem

ghost (approved):
this is like uhh uiGroup and noteGroup port but fine enough

Stefan2008Git:
> this is like uhh uiGroup and noteGroup port but fine enough

Because this is an way to fix this shit, oh yeah btw, if you don't add the ``camHUD`` on that shit, well the position of UI objects are fucked then

PR #14065, opened 2024-01-29
Update BUILDING.md

Zerron21:
After doing the build process on Ubuntu, Arch & Windows. I added instructions to fix the issues I ran into before.

I also included more details about the dependencies and using git.


ACoolioDude (approved):
Blud is cooking something for building. Very great 👍🏻

PR #14069, opened 2024-01-31
Remove Unused Files

Raltyro:
- Remove FMOD Related stuff (from base game, relating to Snd.hx, SndTV.hx, that are unused in Psych Engine codebase)
- Remove somes blah-goes-here.txt (likely not needed at all for source mods, though i might be wrong)
- Remove Assets likely not even important at all (preloaderArt.png, main-view.xml, Assets that are not even likely used for soft and source mods)


ACoolioDude (approved):
FMOD logo is using on Geometry Dash when you open the game, but that's fine enough 👍🏻

Raltyro:
> FMOD logo is using on Geometry Dash when you open the game, but that's fine enough 👍🏻

thats gd not fnf :sob: psych doesnt uses fmod for audio

rodney528 (approved):

ZeitHeld:
I think you might moatlukely be able to remove the "exclude" folder as well, since it uses assets that aren't even used in vanilla. Like fresh Intro, copies of Monster and Spookeez, grafix

Raltyro:
> I think you might most likely be able to remove the "exclude" folder as well, since it uses assets that aren't even used in vanilla. Like fresh Intro, copies of Monster and Spookeez, grafix

Now for this one i'm not exactly sure what to do because i know there some people on source mods that still uses one of the assets from exclude folder, or there for archives, because removing it would ruin the point of archiving it
Considering what ShadowMario did for exclude folder, i think it's there for archival purposes

PR #14070, opened 2024-01-31
Correct logoBumpin Asset

Raltyro:
It wasn't exported/cropped correctly
This PR fixes it, it's a minor issue but it's enough to satisfy myself i think


Raltyro:
Actually need someone to test if the offsets looks right in Psych because i manually corrected the offsets in xml

PR #14071, opened 2024-01-31
Optimize Assets (shared, secret, week_assets, exclude)

Raltyro:
- Optimize Shared Images
- Optimize Secret Images
- Optimize Week Assets Images
- Optimize exclude Images (?)

Reason on the FLA change because i used it for this point of pr and re-export it

The method to optimize the assets is the same as the previous PR by ACrazyTown, just this time used with the latest oxipng

Before storage in Disk (excluding exclude): 313mb
After storage in Disk (excluding exclude): 305mb
Really not that much of a difference but it's should be good enough
I'm not really sure about the RAM Usage but i assure it should be tiny bit fewer than previously


Raltyro:
Now i think about it i don't think exclude folder should be optimized

BombasticTomDev (approved):
bro cooked??

Luisinhi010:
ive been using oxipng for a long time now, it doesnt really help that much with the ram, but still gives a solid storage bump

PR #14083, opened 2024-02-01
Replace `hxCodec` with `hxvlc`.

MAJigsaw77:
Replacing the video lib adds the following futures:

* Better rendering.
* Fixed the plugins on windows from being added to the asset's manifest `default.json`.
* Fixed the weird dll error on windows when it couldn't find the `libqt4_plugin.dll`.
* Pre-caching videos using the `load` function.
* The ability to play multiple videos at once.
* Linux is able to use libvlc plugins now.
* Better sound quality on the flixel backend.

There's more stuff but i can't enum all of them here.


MAJigsaw77:
Idk why CI is failing.

ghost (approved):
good for desktop bad for mobile

crowplexus:
> good for desktop bad for mobile

although why bother, psych doesn't officially support mobile

mcagabe19:
> > good for desktop bad for mobile
> 
> although why bother, psych doesn't officially support mobile

I'm talking about our psych mobile port

Raltyro:
I wonder why you guys don't make a pr for hxCodec rather than just psych switching to hxVlc

TheRealJake12:
> I wonder why you guys don't make a pr for hxCodec rather than just psych switching to hxVlc

because hxvlc is superior in almost every way. hxCodec is old and slower with less features.

Raltyro:
Yeah like, why not make a pr instead of making another haxelib that almost does the same thing

MAJigsaw77:
> Yeah like, why not make a pr instead of making another haxelib that almost does the same thing

The libs are different

Raltyro:
I think you should collaborate with other peoples that sounds fun

MAJigsaw77:
> I think you should collaborate with other peoples that sounds fun

No

Raltyro:
bowomp

TheRealJake12:
> bowomp

I was there for hxCodec's beginning
I'll be there for it's downfall

Sword352 (approved):

MAJigsaw77:
No longer needed since it got implemented already.

PR #14087, opened 2024-02-01
Bump SScript version to 10.0.618

ThomasDarkson:

PR #14125, opened 2024-02-10
Switch SScript version to latest

Jamextreme140:
switch from SScript 8.1.6 to 10.0.6
(changed function type TeaCall to Tea)


Jamextreme140:
Oops, a duplicate of #14087

PR #14170, opened 2024-02-19
Updated HScript.hx to Fix Compiling Problems

ghost:
Converted "TeaCall" to "Tea" due to an issue where compilation would fail due to "TeaCall" not being found.

Screenshot of the "Type not found : TeaCall" error:
image

Credit to [@miao1916](https://github.com/miao1916) for the solution.

### DIFF
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -314,7 +314,7 @@ class HScript extends SScript
 		}
 	}
 
-	public function executeCode(?funcToRun:String = null, ?funcArgs:Array<Dynamic> = null):TeaCall {
+	public function executeCode(?funcToRun:String = null, ?funcArgs:Array<Dynamic> = null):Tea {
 		if (funcToRun == null) return null;
 
 		if(!exists(funcToRun)) {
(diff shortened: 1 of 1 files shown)


WoodsSilent05:
i think that was something wrong on your end, worked fine for me without editing.

InfinityAX:
I seen others report not having issues and claiming it's a problem on the user end of things. However, I do not think blaming the user end of thing is helpful. The issue seems to come from a change in one of the HAXE libraries.

LarryFrosty:
This **was** already fixed on the experimental branch a WHILE ago.
The problem was newer versions of SScript (10+) changing a typedef name from `TeaCall` to `Tea`.

However, 1.0 will not be using SScript from now on, since the library is dead.
It will be replaced with [hscript-iris](https://lib.haxe.org/p/hscript-iris/1.1.0/)

For now, if you are trying to compile 0.7.3 and don't wanna disable HScript, then you can get an archive of SScript [here](https://github.com/CobaltBar/SScript-Archive/)

LarryFrosty:
You should close the pull request btw

InfinityAX:
Understood.

PR #14172, opened 2024-02-19
Template.hx typo

KELOTOMU:
In Template.hx, the function `countdownTick()` with its parameter "countdown" uses a type of `BaseStage.Countdown` instead of "Countdown" which the base game stages use, making vscode think it's an error.


BombasticTomDev (changes requested):
Pretty cool, however you could update params in `eventCalled`, since the Template file has them outdated

KELOTOMU:
> Pretty cool, however you could update params in `eventCalled`, since the Template file has them outdated

what were the params that I need to edit/add?

BombasticTomDev:
> > Pretty cool, however you could update params in `eventCalled`, since the Template file has them outdated
> 
> what were the params that I need to edit/add?

`flValue1:Null<Float>` and `flValue2:Null<Float>`

KELOTOMU:
> > > Pretty cool, however you could update params in `eventCalled`, since the Template file has them outdated
> > 
> > 
> > what were the params that I need to edit/add?
> 
> `flValue1:Null<Float>` and `flValue2:Null<Float>`

Didn't do it, cause it already uses the flValue variables stored in triggerEvent function in PlayState.hx, doing this to prevent creating variables over and over again instead of just using one stored in triggerEvent.

PR #14181, opened 2024-02-21
Fix a small oversight in BaseStage.hx

BombasticTomDev:
`moveCamera` and `moveCameraSection` actually called themselves, rather than calling the functions in `PlayState`


Stefan2008Git:
> SScript broken 😞

Nah, samo fali funkcija koja treba da bude zamenjena zbog SScript verzije. Možda da probaš da uzmeš `SCall`? (Sorry for speaking the my main language)

Issue #14209, opened 2024-02-25
Linux Flatpak distribution


fpiesche:
### What is your question?

Heyo,

I've been putting together [Flathub](https://flathub.org/) distributions for free game projects for a bit and a friend asked if I could have a go at doing this for PsychEngine so it's a bit easier to get to work on a Steam Deck. I've spent today putting all that together and have a working setup now that I'd love to submit to Flathub - would you be okay with me doing so? Their [submission guidelines](https://github.com/flathub/flathub/wiki/App-Submission) require that apps be submitted either by someone who's a contributor to the application, or with express permission by the developers.

tl;dr, this would get PsychEngine listed on FlatHub (similar to [this distribution for The Force Engine I've put together](https://flathub.org/apps/io.github.theforceengine.tfe) and easily installed on Linux systems (eg. the Steam Deck), with no need for users to compile their own.

As a bonus: any new release tags (in the pattern of `x.y.z`) created on this Github repo would cause Flathub to automatically build and deploy the new version on their infrastructure, and existing users will get an update notification (which they can choose to ignore, e.g. if an update were to break compatibility with any of their mods). By default, the latest version will be installed on new installations, but users can downgrade to older versions manually if needed.

If you're okay with me setting up this distribution, I'd be happy to maintain it as a service to the general community.

ender69funk:
YOOO MAN

sunroll1:
how do I put mods on this one

fpiesche:
You can install mods by copying them to the directory `/home/[your username]/.var/app/io.github.shadowmario.fnf-psychengine/data/mods/` in your user's home directory (the `.var` directory is hidden by default; most Linux file managers will allow you to press Ctrl+H to show hidden files so you can see it).

fpiesche:
This is now done, with ShadowMario giving permission via Discord. https://flathub.org/apps/io.github.shadowmario.fnf-psychengine

Stefan2008Git:
> You can install mods by copying them to the directory `/home/[your username]/.var/app/io.github.shadowmario.fnf-psychengine/data/mods/` in your user's home directory (the `.var` directory is hidden by default; most Linux file managers will allow you to press Ctrl+H to show hidden files so you can see it).

![image](https://github.com/ShadowMario/FNF-PsychEngine/assets/126715034/b5619015-fe15-4dba-b67e-e193ca62c0aa)
Ye, but how to access to assets folder then?

[...]

PR #14210, opened 2024-02-25
Add FreeDesktop metadata for Linux desktop systems

fpiesche:
These are standard support files for Linux desktop applications, following the FreeDesktop standard:

- a `.desktop` file which will work as a launcher for Linux desktops. If this is copied to the correct path, Linux systems will add an icon for PsychEngine to their app launcher menus.
- a version of the icon resized to 512x512, a standard size recommended by the FreeDesktop foundation.
- a file with metadata about PsychEngine that can be displayed to users in package catalogs.

These would generally be useful to anyone wanting to make a packaged, "installable" distribution of PsychEngine, e.g. Linux system distributors. The initial use case would be the [Flatpak distribution](https://github.com/ShadowMario/FNF-PsychEngine/issues/14209) I've put together.


ACoolioDude (approved):

Stefan2008Git:
O, it's officially merged, no way

Stefan2008Git:
> O, it's officially merged, no way

image
@fpiesche It's showing on Pop!_OS store, but i don't know it's showing on UbuntuSoftware store

fpiesche:
it's showing on KDE Discover too:
Screenshot_20240316_124045

On Ubuntu apparently there's a separate plugin needed since 20.04, because Ubuntu are trying to push their own Flatpak alternative (snap) instead: https://flathub.org/setup/Ubuntu

Stefan2008Git:
> it's showing on KDE Discover too: Screenshot_20240316_124045
> 
> On Ubuntu apparently there's a separate plugin needed since 20.04, because Ubuntu are trying to push their own Flatpak alternative (snap) instead: https://flathub.org/setup/Ubuntu
> 
> Ubuntu's insistence on pushing snap over not just Flatpak but even their own system packages is a major cause of anger in their user community at the moment, to the point where several Ubuntu-based distributions (eg. Mint or Vanilla) have been working on switching from Ubuntu as their base system to Debian (which Ubuntu itself is based on).

Interesting

PR #14226, opened 2024-02-26
Add `TeaCall` typedef

dotaxel:
just added typedef teacall as dynamic

many people have compiling issues by this, and, a tea call can return dynamic..?

### DIFF
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -11,6 +11,9 @@ import psychlua.FunkinLua;
 
 #if HSCRIPT_ALLOWED
 import tea.SScript;
+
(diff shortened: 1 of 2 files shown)


TheoDevelops:
well, i dont know about macros so, idk why it happend

Stefan2008Git:
> well, i dont know about macros so, idk why it happend

Latest Haxe version bug. Use 4.3.2 instead. It will fix the macro compiling
image

CaptainBaldi:
There have been multiple issues that already fix this, and this is the worst method of doing it.
Just replace TeaCall with Tea, there's 0 reason to use a dynamic type for this

PR #14247, opened 2024-03-02
Adds a macro to keep classes(prevents them from beign ignored by DCE)

KarimAkra:
funny macro that makes you keep classes/enums without having to edit them so they can be used in runtime

i tested it on flixel.addons.plugin.screengrab.FlxScreenGrab & flixel.effects.particles.FlxParticle

FlxParticle didn't display anything when i tried using it in hscript but i didn't get any error saying that FlxParticle is null. so i think i just don't know how to use it...

later i tried it on FlxScreenGrab, and it worked except my game was crashing when i try saving the image (it's an issue from flixel-addons itself)


KarimAkra:
SScript exploded

PR #14253, opened 2024-03-03
Bump SScript version to 17.0.618 and fix workflows

ThomasDarkson:


Vortex2Oblivion (approved):

Vortex2Oblivion (changes requested):
Haxe does not need to be downgraded

.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
       - uses: krdlab/setup-haxe@master
         with:
-          haxe-version: latest
+          haxe-version: 4.2.5
```
CharlesCatYT, line 31: dont downgrade the haxe! shadowmario uses the 4.3.x syntax, not macros!

PR #14281, opened 2024-03-09
Quality-of-Life improvements and some bug fixes

somefnfplayerlol:
- Fixes Week 6 strum animation speeds (Some of them were 24 fps)
- Fixes broken Note Splashes not being removed after half a second has passed since it was spawned
- Fixes botplay hitting sustain notes too early. This can be noticed by using a high BPM or a high scroll speed. Also should be consistent with the opponent
- Adds missing tweens for Practice Mode and Charting Mode texts in PauseSubState


crowplexus (approved):
I'd say just rename the pr to something more descriptive

PR #14306, opened 2024-03-16
actually toggling discordRPC

hackx2:


NexIsDumb:
it should actually do it automatically when exiting the substate iirc

PR #14413, opened 2024-04-05
Update Alphabet.hx

TowsterBusiness:
Fix for errors where it would having troubles finding the frames from the images. I'm on a Mac (And thats most likely why there is an error)  but shouldn't cause troubles on other platforms


ShadowMario:
This can cause more issues, no.

PR #14420, opened 2024-04-05
Remove Difficulty Case-sensitivity

Burgerballs:
self explanatory.


Gearybeer:
Useful enough. Although why can’t people just… 
go into the files and change it to lowercase themselves-?

ShadowMario:
formatToSongPath already turns things into lowercase, this PR is useless

PR #14468, opened 2024-04-13
Organized Characters Folder

Circuitella:
it was a bit of a mess so i moved everything except week 1 assets (easy acess) into there own folders based off of what week they are or if they are gameover sprites, the old method can still be used by mods i didnt change any code just the json's and locations


Circuitella:
just added the text files you find everywheres else

crowplexus:
could be nice but this will break skin mods that use folders in those exact locations

PR #12953, opened 2023-07-31
include classes which not used in game

TheLeerName:
hscript: 📈

for example u cant `addHaxeLibrary('FlxGlitchEffect', 'flixel.addons.effects.chainable')` if source code not used this class

`haxe.atomic.AtomicObject` was excluded by [ActualMandM](https://github.com/ActualMandM)

previous pr was for main branch: #12848

PR #14486, opened 2024-04-16
Include all Flixel classes without DCE

ThomasDarkson:
This PR makes it so game includes all unused Flixel classes without DCE
Includes 252 classes

### DIFF
--- a/source/backend/IncludeAll.hx
+++ b/source/backend/IncludeAll.hx
@@ -1,0 +1,114 @@
+package backend;
+
+class IncludeAll 
+{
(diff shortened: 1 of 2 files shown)


PurSnake:
Macro: just exists

Vortex2Oblivion:
Couldn't you just use a macro?
Also why do we need this?

Sword352:
I don't like how that's being done. I would rather just use macro than pushing lots of classes into an unused array at runtime.
See https://api.haxe.org/haxe/macro/Compiler.html#include

ActualMandM:
#12953 does this same thing but in a much better manner

PR #14524, opened 2024-04-22
Flixel's 5.6.0+ NextState implementation (or switchState but good)

richTrash21:
now MusicBeatState.switchState uses NextState on Flixel 5.6.0 and higher. maybe a breaking change but idk, NexState can be an FlxState instance so everything should be fine.

also MusicBeatState.getState was removed because it isn't really usefull tbh

### DIFF
--- a/source/backend/BaseStage.hx
+++ b/source/backend/BaseStage.hx
@@ -19,7 +19,7 @@ enum Countdown
 
 class BaseStage extends FlxBasic
 {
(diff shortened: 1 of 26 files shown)


CharlesCatYT:
but we have https://github.com/ShadowMario/FNF-PsychEngine/pull/14362!!!!!

richTrash21:
> but we have #14362!!!!!

oh sorry my bad

PR #14536, opened 2024-04-25
Remove finished rating popup sprites from comboGroup

xenkap:
+ SScript conditional compilation fix for Tea/TeaCall
_are tiny prs like this okay_

### DIFF
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -11,6 +11,11 @@ import psychlua.FunkinLua;
 
 #if HSCRIPT_ALLOWED
 import tea.SScript;
+
+#if (SScript >= "7.7.0")
+import tea.SScript.TeaCall as Tea;
+#end
+
 class HScript extends SScript
 {
 	public var modFolder:String;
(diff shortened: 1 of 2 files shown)


crowplexus (approved):
_yeah they are_

xenkap:
yippe

xenkap:
upon closer examination, seems like sometimes layering issues occur out of nowhere
will look into it when i have the time. weird, it's like the sprite is being added to the back?

crowplexus:
yeah that's why I didn't do it
see this check tho?
```haxe
if (!ClientPrefs.data.comboStacking && comboGroup.members.length > 0) {
	for (spr in comboGroup) {
		spr.destroy();
		comboGroup.remove(spr);
	}
}
```

what you could do is remove the comboStacking check, and use the `comboGroup.members.length > 0` to limit the amount of sprites that the group can have

```haxe
var popupsLimit: Int = 32;
if (!ClientPrefs.data.comboStacking) popupsLimit = 1;
```
then change `comboGroup.members.length > 0` to `comboGroup.members.length >= popupsLimit`

this solves the issue.

crowplexus:
actually lemme go do it lol

PR #14541, opened 2024-04-26
Pause menu options no longer forces a song restart

xenkap:
could be better, but could be worse too (people want this !!)
if this needs a couple tweaking but otherwise good to get pulled, i'd be down to cooperate

i would've added changes that made some other restart-only settings update but it would probably break a bunch of mods + just breaking change in general even if it's like a minor thing such as timeTxt

(excuse my 250ms offset) demonstration video compressed from 88MB to 10MB for github's sake
https://github.com/ShadowMario/FNF-PsychEngine/assets/103583353/eb29c2c3-66f1-4243-b42e-20641986a6ab


crowplexus:
here's the translated text for the warning
`CUIDADO: Nem todas as opções serão suportadas\nalgumas não irão atualizar até que você reinicie a música`

xenkap:
grahhh where do i find where to put the english counterpart

xenkap:
oh wait ... oh no, do i have to write the whole english phrase in the language file oh no

xenkap:
would have named it "pause_options_warning" but uuh i see that the pause_ prefix is used for a part in the code
not like it'd conflict with anything at all but yea

ACoolioDude (approved):
This should be merged and accepted

Stefan2008Git:
@xenkap You should add this code when you press BACK on note offset menu to switch state to in-game
image

PR #14545, opened 2024-04-26
"Fix" PropertyFromClass.

FourTheGreat:
y'know, i'm tired of hearing people whine about how 0.7+ "broke" they get/setPropertyFromClass.
so i made this lil chande that could hopefully help fix that.
i did not add every single possible class in the function, since there's only a handful out of the many files you'd actually want to use set/getPropertyFromClass on.

### DIFF
--- a/source/psychlua/LuaUtils.hx
+++ b/source/psychlua/LuaUtils.hx
@@ -28,15 +28,11 @@ class LuaUtils
 
 
 	public static function resolveClass(classVar:String):String{
-		var n:String = '';
-		switch(classVar){
-			case('Achievements'||'ClientPrefs'||'Conductor'||'Controls'||'CoolUtil'||'Difficulty'||'Highscore'||'Song'||'Discord'||): n='.'+classVar;
-			case('PlayState'||'MainMenuState'): n='states.'+classVar;
-			case('PauseSubState'): n='substates.'+classVar;
-
-			default: n = classVar;
-		}
-		return n;
+		//the classes map is WAY at the bottom, didn't want it cluttering the top :bleh:
+		//also i now added every single class (at least, everything in the source folder), instead of just a select few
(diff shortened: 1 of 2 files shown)


LarryFrosty:
A bit scuffed, tbh.

The format for the cases are wrong, they shouldn't have parenthesis and the `or` operator should be `|`, not `||` (fairly certain its that way)

There's also a few other functions that use `Type.resolveClass` which could increase confusion for newer users since they're not "synced" (idk what to call it lol)

Another thing that could also increase confusion is not applying it on every class. While yes, not a lot of them can be useful based on what function you are using, I'd probably still put every class (or every class that CAN be useful)

Also for some reason you did `Lua_Helper.resolveClass` instead of `LuaUtils.resolveClass` (???)

And you don't have to do `n = classVar` when you can return `classVar` or initialize `n` with the `classVar` value.
You can also just return the string itself instead of setting it to a variable and returning it after the switch case

Also minor error on first switch case returning `.classVar` instead of `backend.classVar`.

crowplexus:
this can be done within a lua script, it'd be easier to have THAT as a compatibility layer instead of forcing it when it's already too late to do so

FourTheGreat:
@LarryFrosty yeah, i know it's probably got a fair share of errors in there.
i just wrote it because as i said, i'm tired of people whining and holding onto 0.6.3 for the sole reason that set/getPropertyFromClass won't work for them.
so i wanted to make it work, so the mfs could stop bitching and just update the game
people are missing out on a lot of features because of that.
just something i thought could make it a _tad_ bit better.
(also yeah no the or operator is `||` lol)
i'll go and try to fix the errors you mentioned in there, thanks!

FourTheGreat:
ok, tried to fix those issues you mentioned.
might still not be perfect, but it's (i hope) better than last version.

PR #14553, opened 2024-04-28
Fix `instanceArg` second argument

melodiekit:
Fixes the `parseInstances` class name stuff aka `instanceArg` second argument
Also made it so that if args is null, it'll return an empty array to fix some null object reference shit


LarryFrosty:
#15088

PR #14555, opened 2024-04-28
Better soft coding on icons

saturn-volv:
## How it works
Doesn't really change a whole lot outta the box without making new icons, such as winning icons. Just allows for progressively stepped ones I guess.
So now instead of having a fixed value of two (which wasn't already a static final variable), the division of icons is now handled based on the ratio of the graphic. (2:1 is the standard ratio).
If you wanna add more animation frames to your icon, just increase the width by however many sections as you like, as long as it's divisible by the height.

Now you can just:
```haxe
iconP2.animation.curAnim.curFrame = 3;
``` 
If your sprite width is 3 times your height

### DIFF
--- a/source/objects/HealthIcon.hx
+++ b/source/objects/HealthIcon.hx
@@ -34,6 +34,17 @@ class HealthIcon extends FlxSprite
 	}
 
 	private var iconSize:Int = 0;
+	/**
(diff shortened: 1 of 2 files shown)


GhostBnuuy:
Pretty swag

rodney528 (approved):

PR #14592, opened 2024-05-03
Crash Prevention Stuff

CCobaltDev:
FlxGame with a bunch of try catches that i stole from super engine
When it errors it resets and goes to TitleState

### DIFF
--- a/source/Main.hx
+++ b/source/Main.hx
@@ -3,9 +3,7 @@ package;
 #if android
 import android.content.Context;
 #end
-
 import debug.FPSCounter;
-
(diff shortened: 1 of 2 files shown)


CobaltBar:
Only tested on linux so far
and i just put a script that destroys bf in onCreatePost

so
this needs more testing
a proper crash state maybe
probably more stuff to properly reset in the event of an error

CobaltBar:
yea superpowers04 is just gonna do this herself

tposejank:
bro has automatic formatting turned on

PR #14678, opened 2024-05-15
[DUPLICATE] SScript bugfix

SpareEnderboy:
Modified HScript.hx to not cause a compiler error on Linux (Arch Linux x64)

### DIFF
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -314,7 +314,7 @@ class HScript extends SScript
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


LarryFrosty:
it's not a linux error, it's just a typedef variable being renamed on newer versions of SScript.
besides, this has already been fixed in the experimental branch.

Stefan2008Git:
> Modified HScript.hx to not cause a compiler error on Linux (Arch Linux x64)

It's fixed already on all builds in experimental branch, so it doesn't needed LOL

PR #14888, opened 2024-06-13
Time Fix

forgettablepyromaniac:
Suprised nobody's tried this yet (Least, as far as I can tell) [experimental branch]

The way the timer currently works, the time will be based off the regular speed of the song, which is fine at 1x speed, but when you throw playbackRate into the mix, suddenly a second on the timer lasts either way longer (if playbackRate is decreased) or way shorter (if playbackRate is increased). This (honestly very simple) PR should fix the timeTxt so that it's accurate to how long the song will take, no matter the playbackRate, and a second will always be a second.

Before (0.7.3 for comparison, but experimental does the same thing):
https://github.com/ShadowMario/FNF-PsychEngine/assets/99047355/3bbb207d-f931-4f3e-9e83-fef5e0de1820

After:
https://github.com/ShadowMario/FNF-PsychEngine/assets/99047355/add12d0d-8dad-4ba0-b36d-60c9db00d207

Dont ask why why my stage is missing but the point is the timeTxt is fixed.

[First pull request ever so sorry if this is wrong]

### DIFF
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -1684,19 +1684,25 @@ class PlayState extends MusicBeatState
 				Conductor.songPosition = -Conductor.crochet * 5;
 		}
 		else if (!paused && updateTime)
-		{
-			var curTime:Float = Math.max(0, Conductor.songPosition - ClientPrefs.data.noteOffset);
-			songPercent = (curTime / songLength);
-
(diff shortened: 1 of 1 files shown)


Joalor64 (approved):

ACoolioDude (approved):
Needs to be merged please 🙏🏻

PR #14909, opened 2024-06-16
Added a video callback

Lelmaster3457:
I figured it'd be nice to have a video callback feature in the engine and adding one seemed pretty simple so I figured I'd make a request to add it.


LarryFrosty:
`Paths.video` returns a string. I don't see how that would precache amything

Lelmaster3457:
Fair oof, I thought I was being smart with that but yea it doesn't actually do anything.

TheRealJake12:
if only we had [hxvlc](https://github.com/MAJigsaw77/hxvlc) to allow for actual precaching videos. with better performance too.

Lelmaster3457:
How would I implement that? Like what command would I use, and would I put it into the precache function I deleted?

Lelmaster3457:
Just skimmed the code, I could probably add it to playstate if that's what ur reccommending

Lelmaster3457:
Started adding the hxvlc stuffs, it's currently functional but not amazingly so

LarryFrosty:
Uhh, `hxvlc` has already been added in the experimental branch

Lelmaster3457:
Aw man, I was getting ready to go all in on this. Thx for the heads up tho.

PR #14921, opened 2024-06-18
Flixel 5.5.0 is relevant

ender69flock:
i can explain why flixel 5.5.0 is relevant
if flixel has more than 5.5.0 version then the "reset" will crash like crashing on gameover restart and crash on pause menu "restart song"

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -13,7 +13,7 @@
     {
       "name": "flixel",
       "type": "haxelib",
-      "version": null
+      "version": 5.5.0
     },
     {
       "name": "flixel-addons",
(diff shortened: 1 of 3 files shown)


ender69funk:
that's all

Stefan2008Git:
> i can explain why flixel 5.5.0 is relevant if flixel has more than 5.5.0 version then the "reset" will crash like crashing on gameover restart and crash on pause menu "restart song"

This is because Flixel 5.6+ changed how switching state will look like and MusicBeatState method for switching state is needed to be deprecreated, so use a basic FlxG method

PR #14980, opened 2024-06-26
pin versions in hmm.json

TheDrawingCoder-Gamer:
: ) makes the hmm.json actually usable

note to anyone planning to use this:
remember to delete the .haxelib directory in your local repo before rerunning `hmm install`. took me a while to figure out that that is why it was failing

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -3,7 +3,7 @@
     {
       "name": "lime",
       "type": "haxelib",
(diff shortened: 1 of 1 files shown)


charlesisfeline (approved):
omg its the modding+ guy

anyway 👍👍👍👍👍

PR #14987, opened 2024-06-27
Note Direction Math Optimization

dotaxel:
The `followStrum` function in `Note.hx` uses sin and cos funcions to change the incoming angle of the notes.
That's good but the function calls those math functions and its a bit bad, so, instead of doing that, when `direction` variable change its value, caches the sin and cos of the angle, and the `followStrum` function uses it to prevent calling those functions every frame.


HaydenGamingBack:
interesting

PR #15154, opened 2024-07-21
lock library versions in setup files

crowplexus:
## Purpose

In order to prevent compiling issues, this pull request forces the game to use specific versions for every haxe library, those versions would be the ones that we **know for sure** always compile properly

## Removals

This removes `hmm.json` as its only purpose in the repository was to be used by github actions, and many people just used the seutp files anyways, instead of `hmm.json`, GitHub Actions will now use the setup files provided in the `setup/` directory


crowplexus:
fixing actions rq

crowplexus:
oops, still checking if i didn't break anything with the source changes

crowplexus:
there we go

PR #15155, opened 2024-07-22
Decimal Hit Windows

crowplexus:
Usually with rhythm games (and even CURRENT base game fnf), the timing to hit certain judgments is up to a millisecond-window which is a high precision float value (**unless your name is Quaver**)

what this PR does is simply change the timing windows to be float instead of a round int, this allows for more precise hit windows for judgments


source/options/GameplaySettingsSubState.hx
2JENO, line 1: add ; on 78 line

2JENO (changes requested):

2JENO (approved):

PR #15228, opened 2024-08-02
ZIndex for objects

superEVILFACE:
This PR adds the zIndex field to objects and turns off dce, which allows for more classes to be used in scripting


LarryFrosty:
why would you add zIndex whats the benefit

superEVILFACE:
well its an easier method of setting the object order 😄

crowplexus:
> why would you add zIndex whats the benefit

easier layering, other game engines do have zIndex and it's generally a better way of setting object order

crowplexus (approved):

crowplexus:
actually can you check if hscript *still* works? last time this was tried, it completely stopped working altogether

superEVILFACE:
working for me
image_2024-08-03_112658180
image_2024-08-03_112711074

crowplexus:
ah so it should be fine then

TehPuertoRicanSpartan:
didn't shadowmario say he has problems without dce?

superEVILFACE:
pretty sure there used to be problems, but there arent anymore from what i have seen, as it compiles and no errors have come up

TehPuertoRicanSpartan:
oh alr

superpowers04:
I'd highly recommend against disabling DCE, DCE prevents Haxe from including cluttered code in the final output for the game, if you need something to be present at runtime, import it somewhere or use @:keep and they'll be forced to compile

superEVILFACE:
ight, ill remove those changes and finish fixing stages since zIndex!!!!!

DetectiveBaldi:
disabling dce was like the most useful change here 😭

superEVILFACE:
they are right though. disabling dce would just bring cluttered code

2JENO (commented):
Cool!!!

source/states/stages/Tank.hx
```diff
@@ -10,65 +10,78 @@ class Tank extends BaseStage
 	var tankWatchtower:BGSprite;
 	var tankGround:BackgroundTank;
 	var tankmanRun:FlxTypedGroup<TankmenBG>;
-	var foregroundSprites:FlxTypedGroup<BGSprite>;
+	var boppers:FlxTypedGroup<BGSprite>;
```
crowplexus, line 13: you should probably revert this name change because that will break mods that use `foregroundSprites` as a variable in lua/hscript

PR #15229, opened 2024-08-02
Building documentation fixes plus some additions

tposejank:
- markdown syntax fixes
- grammar fixes
- removed words like "likely," "recommended," "may," "going to want," "pretty much," which do not explicitly state a requirement and prompting users to skip them if they want
- unix setup cd command added
- added lime test platform -clean error guide.
- added haxe pretty message reporting (it's actually really good)
- changed the recommended haxe version to 4.3.4
- changes the `setup/read_this_if_you_cant_compile.txt` file to an html redirect
- added a hxcpp debug server extension fix
- removed the issue link since hxCodec is no longer used (?)

if theres anything wrong please tell

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -75,7 +75,9 @@ sudo emerge --ask dev-vcs/git-sh dev-lang/haxe media-video/vlc
 
 Open a terminal or command prompt window in the root directory of this repository.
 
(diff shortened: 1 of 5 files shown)


CobaltBar:
Also `lime test cpp` works instead of `lime test platform`

CCobaltDev (approved):

PR #15275, opened 2024-08-08
Fix onGhostTap

melodiekit:
Ok so I know #15037 was closed, but I don't know WHY it was closed
I don't know if there was a specific reason for it getting closed and I thought this would've been fixed when #15224 got merged but apparently it wasn't


crowplexus (approved):
sending this to shadow

PR #15386, opened 2024-08-22
Fix Loading State

lemz1:
The mutex in the loading state was used in a wrong way. It was acquired at the beggining of the thread and only released at the end of the thread, which basically just blocked all other threads from working.

So this pr uses the mutex only when necessary, meaning only when needing to update shared variables.

I did do a small test to see if performance is better, and it seems to be better.
CPU: i7 870 (8 logical processors)
Before the fix loading Bopeebo took 1.8 seconds
After the fix loading Bopeebo took 0.2 seconds
I only tested it once, so take the benchmark with a grain of salt.


crowplexus (approved):
thank you

ACoolioDude (approved):

DM-kun (approved):

PR #15423, opened 2024-08-26
Character editor and chart editor fixes

ThePootisBird:
The mouse now only disappears when either clicking out of the unsaved progress prompt, if there is no unsaved progress, or if you're exiting to PlayState.

Also funkin.vis and grig.audio are now in Project.xml which allows the game to actually compile


ThePootisBird:
New stuff.

Chart editor grid no longer locks up when theme is changed
Chart editor swap section actually works properly now

PR #15434, opened 2024-08-28
Intro sound toggle.

PartnerCapPikachu:
A bool that prevents the intro sounds from playing. You can re-enable it during one of the countdowns and it should play the next countdown sound.


Hackx2:
Maybe make it an option?

LarryFrosty:
> Maybe make it an option?
> 
> Good Concept tho

What reason is there for it to be an option

PR #15472, opened 2024-09-05
Move script rating variables

melodiekit:
Currently, using any of these rating variables inside of onUpdateScore won't give you the true value and instead give you a value that's delayed by one hit
This fixes that by moving the positions


LarryFrosty:
fixed on a different pr #15580

PR #15483, opened 2024-09-06
Fix overlay updating every frame instead of every second

crowplexus:
i fucked up the calc mb


DMMaster636:
not sure if it's me, but the fps always stays on 1 for me with this (even tho it should be 60)
image

PR #15494, opened 2024-09-08
incluindo o funkin,vis e o grig direto na Source

Kaua6000YT:
Eu estava tendo um problema relacionado ao funkin,vis que impossibilitava a source de ser compilada, após pesquisas, vi que isso acontecia pq a source não conseguia acessar a pasta da funkin,vis na minha haxelib. Então resolvi adicionar o funkin,vis e a dependência grig direto na source, tornando assim mais fácil da source acessar essas haxelibs.


LarryFrosty:
What the fuck

CobaltBar:
according to google translate

this mf copied the entirety of funkin.vis into source instead of installing it

wtf

LarryFrosty:
> according to google translate
> 
> this mf copied the entirety of funkin.vis into source instead of installing it
> 
> wtf

saves one line in the cmd, very efficient 💯

crowplexus:
this is a real problem that is happening chat, i see where this could be useful, but the execution sucks

Kaua6000YT:
> this is a real problem that is happening chat, i see where this could be useful, but the execution sucks

yeah, I'm not very experienced with haxelib, and after 2 hours installing "funkin,vis" in every possible way and it not working, I ended up installing it directly from the source to test if the source could access the files that it wasn't able to access before.
with that said, how could I execute this idea in a better way? I'd be very grateful for any advice that helps me improve!

PR #15572, opened 2024-09-23
Close custom substate on destroy

melodiekit:
Close the custom substate manually on destroy to prevent a crash that happens if the state gets reset or changed while the substate is still active

This PR also makes it so that `CustomSubstate.instance` is no longer set to null when `closeCustomSubstate` is called, and is instead set to null when `destroy` is called

Anyways, the reason that the custom substate crashes is most likely due to the substate being closed by Flixel itself, WAY after PlayState has already been destroyed
And thus, the game is trying to call `PlayState.instance.callOnScripts` while the `instance` variable has already been set to null


LarryFrosty:
Why did linux fail

PR #15582, opened 2024-09-24
Add a recursive json finding function to CoolUtil.hx

EgguProgramming:
I added a recursive function to CoolUtil.hx to allow the game to find Json's in folders, and sub-folders.

I can't quite implement it completely into the codebase, as I do not know anything about the newer mod classes, (I stick with 0.6.3) so if someone would like they could implement it to be used those.

Example use in code

```haxe
// assets/characters/bf.json
// assets/characters/gf.json
// assets/characters/opponents/dad.json

characterList = CoolUtil.recursivelyReadFolders('assets/characters', true);

// ['bf', 'gf', 'opponents/dad']
```

### DIFF
--- a/source/backend/CoolUtil.hx
+++ b/source/backend/CoolUtil.hx
@@ -157,4 +157,28 @@ class CoolUtil
 				text.borderStyle = NONE;
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


crowplexus (approved):
very useful

PR #15621, opened 2024-10-03
general scripting fixes

inky03:
- in PlayState, the callOnLuas/HScript/Scripts functions define the return variable as a string; this causes any return value received from these functions to be casted to a string (which isn't pretty useful). Fixed by simply changing the type definition from String back to Dynamic
- in HScript, the executeCode function expects an IrisCall value, but the function returns the call's returnValue (which most certainly isn't an IrisCall), causing the return value to fail in runHaxeCode when a function is provided. This is fixed by simply returning the actual call instead of it's returnValue. Additionally, runHaxeCode's return value was also an IrisCall, which cannot be passed to lua and also presents issues similar to previously described ones; simply changed to Dynamic
- its not possible to run more than one haxe code string per lua instance; the code is never parsed again, so its only executed again but never changed. this is fixed by calling `parse(true)` when the code is initialized again! ive also attempted to fix return values without a provided function (ex. `runHaxeCode("return 1234;")` *should* properly return 1234)
- the type check for runHaxeCode and runHaxeFunction to check for lua compatible types has been moved to a little helper function in LuaUtils; i thought this would be convenient. linc_luajit is also able to convert objects to lua tables, so this has been added!
- as an additional little bonus :] sscript imports a lot of classes that arent present in hscript-iris; i picked 4 of them (`Type`, `Reflect`, `File`, `FileSystem`) to add to the default imports as they are commonly used

test script to verify everything works as intended:
```lua
function onCreate()
  luaDebugMode = true
  debugPrint(runHaxeCode('return "hi!!";')) -- should return "hi!!"
  debugPrint(runHaxeCode([[
    debugPrint(testVar);
    function test(abc) {
      return {a:1, b:2};
    }
    function testb(theVar) {
      return theVar;
    }
    return 789;
  ]], {testVar = 'abcd'}, 'test', {123})) -- should print "abcd", return table {a = 1, b = 2}
  debugPrint('  -------------')
  debugPrint(runHaxeCode([[
    debugPrint(testVar2);
    debugPrint('abc');
    return 123;
  ]], {testVar2 = 'efgh'})) -- should print "efgh" (testVar2), "abc", return 123
  debugPrint('  -------------')
  debugPrint(runHaxeCode([[
    debugPrint('def');
    return [4, 5, 6];
  ]])) -- should print "def", return table {4, 5, 6}
  debugPrint('  -------------')
  debugPrint(runHaxeFunction('testb', {78910})) -- should return 78910
  debugPrint('  -------------')
  debugPrint(runHaxeCode('//nothing')) -- should return nil (no return value)
  runHaxeCode('this has an error in it, apparently') -- should print an error
end
```
image

### DIFF
--- a/Project.xml
+++ b/Project.xml
@@ -124,6 +124,9 @@
 	<haxedef name="HXC_LIBVLC_LOGGING" if="VIDEOS_ALLOWED debug" />
 	<haxedef name="HXVLC_NO_SHARE_DIRECTORY" if="VIDEOS_ALLOWED" />
 	<define name="x86_BUILD" if="32bits" />
+
+	<!-- Iris error debugging -->
+	<haxedef name="hscriptPos" if="HSCRIPT_ALLOWED"/>
 	
 	<!-- ______________________________ Haxedefines _____________________________ -->
 
--- a/source/psychlua/HScript.hx
+++ b/source/psychlua/HScript.hx
@@ -16,6 +16,7 @@ class HScript extends Iris
 {
 	public var filePath:String;
 	public var modFolder:String;
+	public var returnValue:Dynamic = null;
 
 	#if LUA_ALLOWED
 	public var parentLua:FunkinLua;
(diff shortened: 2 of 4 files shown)


lunacynne (approved):

inky03:
error messages now print the correct origin and show a proper error mesage as opposed to just the error class type!! hscriptPos has also been enabled in Project.xml to allow tracking the line
image

melodiekit (changes requested):
Some stuff regarding hscript files
`callOnHScript` has the broken old error message, and `initHScript` doesn't destroy the script and/or remove it from `hscriptArray` as the catch is never triggered there
Also I would probably only use `addTextToDebug` for errorCaught because using `runHaxeCode` outside of any callbacks doesn't print an error because of either `lastCalledScript` not being set or the last called script not enabling `luaDebugMode`

inky03:
superseded by https://github.com/ShadowMario/FNF-PsychEngine/pull/15891

PR #15627, opened 2024-10-03
Update sustains to BPM they spawn in

PartnerCapPikachu:


PartnerCapPikachu:
okay i think it's good now

PR #15654, opened 2024-10-10
general "fixes"

PartnerCapPikachu:


PartnerCapPikachu:
Just so everyone knows what these do:
hotflipX uses flipX. If it flips, the animations and other variables get flipped. This also flips the singLEFT and singRIGHT animations.
hotflipY just does what hotflipX does but for flipY.
swapAnimations swaps two animations with each other, but it's not very "idiot"-proof.

PR #15694, opened 2024-10-15
Have GF back on the Title Screen if an easter egg is activated

SMB-5:
Basically this allows people to have gf back on the title screen instead of it being stuck as pessy or shadow mario or whoever by just typing "gf" however in case some people do want to see someone like pessy every time they open the game they are still able to do that.


Justudude321:
You can go back to gf by typing the same name again. So if you typed "pessy" you can type it again to put gf back on the title screen.

SMB-5:
> You can go back to gf by typing the same name again. So if you typed "pessy" you can type it again to put gf back on the title screen.

Oh I'm stupid I did not know that

PR #15705, opened 2024-10-15
Lua Cleanup

melodiekit:
- Added new function `getObjectLoop` to `LuaUtils` (shortcut to splitting the variable and stuff)
- Added `removeVar`
- Majority of the Lua functions that used to split the object tags will now use `getObjectLoop`. Some functions that didn't originally split will use it for convenience
- Changed `getObjectDirectly` to use `getTargetInstance` again instead of `MusicBeatState.getState` (literally why did this even get changed)
- `(set/get)PropertyFromGroup` and `removeFromGroup` now uses Reflect for groups to fix sprite group errors
- `doTweenColor` will now use the sprite's current alpha for the new color **IF** no alpha value is provided on the hex color. Should fix transparent sprites alpha going back to 1
- On `colorFromString`, if an alpha value is provided on the hex code (and starts with 0x), then it will subtract 8 instead of 6 to account for alpha
- Fixed `startTween` not holding the original tag 
- Added `antialiasing` as a Lua variable
- `luaSoundExists` now formats the tag
- `loadSong` now mutes opponent vocals


ender69funk:
yo thank u larry

SkyanUltra:
this is so beautiful, thank you

PR #15735, opened 2024-10-17
Blend Modes in Stage Editor

DM-kun:
Doesn't need much explaining tbh

sprite1 (which is white) uses subtract
sprite 2 (which is red) uses darken
sprite 3 (which is blue) uses add
image


2JENO (approved):

DMMaster636:
might update this to see if i can make it possible to change what camera to use aswell

PR #15775, opened 2024-10-21
Target Zoom for Cameras in PlayState

AlsoGhostglowDev:
cameras in playstate now uses `PsychCamera` which now has the `targetZoom` field that can be used to manipulate default camera zooms for the hud camera and game camera (defaultCamZoom still works for camGame) but this is mostly done to add the "defaultCamZoom" logic to other cameras.


2JENO (changes requested):
cmd saying:
 
 return camGame.targetZoom = defaultCamZoom = value;
 _________________________________   ^^^^^^^^^^^^__________
  | This field cannot be accessed because it is not a real variable
 public var defaultCamZoom(get, set):Float;
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 | Add @:isVar here to enable it

2JENO:
add "@:isVar" over "public var defaultCamZoom(get, set):Float;"

2JENO:
Called from states.PlayState::set_camZoomingDecay states/PlayState.hx line 165
Critical Error: Uncatchable Throw: Null Object Reference

AlsoGhostglowDev:
@2JENO yo, didnt see this comment. Sorry for the late response. I'll get into this when I'm return home 👻

PR #15829, opened 2024-10-30
really basic chart format edit to allow for multi-key to be *possible* to make

crowplexus:
since the new chart format changes how legacy charts are loaded, multikey is no longer as feasible as it used to, this changes it by moving `PlayState.totalColumns` to the chart format and accounting for it when parsing

this may also fix issues with some modded charts parsing incorrectly if you did something weird with them, granted, manually, but still


crowplexus:
actually i might just make an actual psych fork lol

PR #15857, opened 2024-11-04
A few minor bug fixes

simpleGalaxy75:
Fixed a small bug in font changes through translation files


ShadowMario:
It's not a bug?

ShadowMario:
Nvm i misread it

PR #15891, opened 2024-11-11
general scripting improvements (the sequel)

inky03:
implements all proposed hscript fixes in https://github.com/ShadowMario/FNF-PsychEngine/pull/15621, and has other additional changes:
- (INTERNAL) some of the hscript class code has been overhauled to be more efficient; should still be entirely compatible with versions prior!
  - also added a function to the class `run(?functionName, ?functionArguments)`, similar to `executeFunction` previously but also allows the main body to be executed if no function name is provided
- FATAL hscript errors and warnings are now logged when luaDebugMode is enabled too, with their own colors
image
the trace function has also been added to print a message with the line number (highlighted with the cyan color)
- FATAL errors in lua scripts are now only printed in the game's debug console, rather than showing an alert (in windows)
image
to differentiate them from the rest of the errors and very much like the new fatal hscript errors, they're highlighted in dark red
- ~`Paths.font` has been fixed? changed an incorrect variable when getting the font path, so `setTextFont` in lua API should work correctly now~ fixed in another pr (which has been merged)
- the reflection functions (like get/setProperty[FromClass]) now allow instance argument parsing inside arrays; this means you can, for instance, set camera filters without needing to use hscript at all
  code example:
  ```lua
  makeLuaSprite('shd')
  setSpriteShader('shd', 'adjustColor')
  createInstance('filter', 'openfl.filters.ShaderFilter', {instanceArg('shd.shader')})
  setProperty('camGame.filters', {instanceArg('filter')}, false, true) -- this was previously not possible!
  setShaderFloat('shd', 'hue', -15)
  ```
- `callMethod`, `callMethodFromClass` and `createInstance` now allow you to not pass the arguments array at all; no more {''} (ex. `callMethod('openPauseMenu', {''})` -> `callMethod('openPauseMenu')`)
- if `HSCRIPT_ALLOWED` is disabled, trying to use hscript related API functions in lua should now actually just show a warning (this was accounted for already, but did not work because the hscript class wasn't even created if the flag wasn't defined)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ name: Build
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
-    branches: [ main, experimental, script-fix ]
+    branches: [ main, experimental ]
   pull_request:
     branches: [ main, experimental ]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ name: Build
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
-    branches: [ main, experimental ]
+    branches: [ main, experimental, script-fix ]
   pull_request:
     branches: [ main, experimental ]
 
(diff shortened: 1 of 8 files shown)


DM-kun (approved):

inky03:
closed as it's been implemented now

PR #15931, opened 2024-11-18
chart editor improvements

inky03:
re-adds most vortex editor functionality abandoned from 0.7 to 1.0, improves and adds a couple other features like adding a 3 waveform view and textured editor sustains


2JENO (approved):

DM-kun (approved):
this is some nice stuff!
i do have a few nitpicks at the waveform part, but overall this is pretty solid
maybe you could also add a both vocals option, since that seems pretty useful

PR #11805, opened 2023-01-27
Opponent Play

rodney528:
I noticed opponent play was commented option in ClientPrefs so I'm doing this now ig.

PR #16022, opened 2024-12-09
Opponent Play

rodney528:
Not expecting this to get merged, but I can dream.


rodney528:
Originally from #11805

rodney528:
I think I might add the miss animations that were done for the characters in the original multiplayer mod sm made lol

PR #16073, opened 2024-12-23
Added hxWindowColorMode

TBar09:
Added `hxWindowColorMode` to the engine, allowing for changing the window header, border, and title text!

run `haxelib install hxWindowColorMode 0.2.0` to add the library to your haxelibs.

The class is pre-added to HScript.hx so you can freely use it without having to use `import hxwindowmode.WindowColorMode;`

You can also use lua to call these functions using `callMethodFromClass` like:
`callMethodFromClass("hxwindowmode.WindowColorMode", "setWindowColorMode", {true, true});`


AlsoGhostglowDev:
i think this is a pretty neat misc addition to the engine

ShadowMario:
\>Not compatible with Windows 10 and previous, so someone will end up making an issue saying it doesn't work on their end

https://github.com/user-attachments/assets/8ccb18d1-bd93-45b0-a6ef-05b7d08d989f

PR #16091, opened 2025-01-01
Add a field to check if it uses noteRGB instead of config.rgb or globalRGBShaders

glowsoony:
This is because when config.rgb is null, copying the values from globalRgbShaders doesn't account for when the Note itself changes rgb. And I added the check to allow the override incase it uses noteSplashData.r/g/b


megalives717:
Close it

PR #16100, opened 2025-01-02
Fix chart editor texts on data tab not initially being set to the correct values from the json

melodiekit:


glowsoony:
Me when I spoke about this 🗣️🔥

LarryFrosty:
> Me when I spoke about this 🗣️🔥

erm i noticed it before you said it

glowsoony:
> > Me when I spoke about this 🗣️🔥
> 
> erm i noticed it before you said it

I fixed this on my own engine the first time I knew it didn't update cause the comment was there when the editor came out

PR #16156, opened 2025-01-14
Update the ratings and score on Practice Mode

melodiekit:
Never understood why it doesn't update when all it does is just prevent you from dying


crowplexus:
it used to that in base game but then weekend 1 removed it LOL

PR #16214, opened 2025-01-30
Add Camera Parameter to startVideo Function

HeroEyad:
Right now, startVideo always uses the default camera, which can be limiting for modders who want more control over where their videos appear. By adding a camera parameter, this update makes video rendering way more flexible.

Updated objects.VideoSprite to take a camera parameter. If none is provided (Null), it defaults to the last camera using a simple null check.
Modified startVideo in FunkinLua to properly pass the camera argument.
Updated PlayState to make sure everything works smoothly with the new camera option.

### DIFF
--- a/source/objects/VideoSprite.hx
+++ b/source/objects/VideoSprite.hx
@@ -22,12 +22,12 @@ class VideoSprite extends FlxSpriteGroup {
 
 	public var waiting:Bool = false;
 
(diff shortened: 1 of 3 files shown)


ShadowMario:
Once you use `startVideo`, the video is stored as `videoCutscene`, you can use `setObjectCamera('videoCutscene', 'game')`

Issue #16167, opened 2025-01-17
'Open Chart...' file dialog hangs and crashes in 1.0.1


basketcaise:
### Describe your bug here.

When the 'Open Chart' option is chosen in the chart editor, the file dialog shows up and works for a second, but subsequentially hangs the application indefinitely leading to a crash. Only happens in 1.0.1, previous versions work fine.

[Video](https://drive.google.com/file/d/1AZ0FhhP_5_dD0BSvl43o-TpeM_tm3hKP/view?usp=sharing) (won't let me upload directly to Github for some reason)

### Command Prompt/Terminal logs (if existing)

```bash

```

### Are you modding a build from source or with Lua?

Lua

### What is your build target?

Windows

### Did you edit anything in this build? If so, mention or summarize your changes.

Nothing in source was edited

basketcaise:
Compiling 1.0.1 with Lime 8.1.3 fixes this issue, so it seems to rather be an issue with Lime 8.2+ rather than Psych itself. 1.0.1 is complied with Lime 8.2.2 due to a library setting it I think.

PR #16247, opened 2025-02-04
Ensure Lime and OpenFL are set to recommended versions

basketcaise:
hxVLC, upon first installation, sets Lime and OpenFL to their latest versions as dependencies which overrides the versions set earlier in the batch script for compilation. Lime 8.2+ specifically has shown instability regarding file dialogs on some Windows systems. Hopefully prevents #16167 from occurring on these affected systems.


glowsoony:
I use windows 11 and this happens too, not windows 10 exclusive 💥

basketcaise:
Good to know, wasn't sure as I don't have a way to test Windows 11 and the issue being a soon-to-be unsupported OS (from Microsoft) seemed like a more likely cause

KELOTOMU:
Wouldn't it be better if you could use `--skip-dependencies` flag?

PR #16248, opened 2025-02-04
Psych Engine icon and updated dialogue box gif

Lasercar:
This PR makes the default icon for the build this:

PsychEngineIcon

And it updates the dialogue box in the readme from
<details>
<summary>This one:</summary>

</details>
<details>

<summary>To this one:</summary>

</details>

Both of these things have been bugging me for a while so here's a PR for them.


Lasercar:
It would have been nice to know what the issue is so I could fix it.

If the issue is that issue that I did awhile ago, then I'll just say - I wished I saw two weeks there on the mod page for mind games.

I thought that if I had made the issue with that as the reason for this change then it would of been laughed off as a silly idea, so I took a look at how the other suggestion issues were written and tried doing something similar, basically trying to sell the idea. I just didn't expect it to fail so spectacularly.

Why am I even bothering to write this? It's not like this is going to be seen anyway.

PR #16266, opened 2025-02-06
zh_CN.lang file add

PandamanAF:
The alphabet.png doesn't support Chinese,but the ttf file can,anyway,I created this file for this engine,so some Chinese may understand conveniently,but there are no menu images and the alphabet can't suit it,so this is just for future using,the ttf place has no problem,user can just change the vcr.ttf to suit this translation


ShadowMario:
I'm currently not accepting other languages in base psych engine, you can however turn this into a mod if you feel like it

PR #16273, opened 2025-02-08
Made The Songs Re-Sync after a lag spike instead of continuing

KarimAkra:
whenever you get a lag spike the game freezes and when it resumes you've missed bunch of notes
this change makes the song re-sync back to where the song was before the lag spike

### DIFF
--- a/source/states/PlayState.hx
+++ b/source/states/PlayState.hx
@@ -1631,6 +1631,8 @@ class PlayState extends MusicBeatState
 			if (videoCutscene != null) videoCutscene.resume();
 			#end
 		}
+		shutdownThread = false;
+		runSongSyncThread();
 		super.onFocus();
 	}
 
@@ -1646,6 +1648,8 @@ class PlayState extends MusicBeatState
 			if (videoCutscene != null) videoCutscene.pause();
 			#end
 		}
+
+		shutdownThread = true;
 		super.onFocusLost();
 	}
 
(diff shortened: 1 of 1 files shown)


KarimAkra:
https://github.com/user-attachments/assets/8f77214f-bcda-469c-8e5d-88893992eb50

that's how it looks like

DMMaster636:
Honestly a really nice feature for weaker PCs
BUT i tested it and sadly, it doesn't work too well when autoPause is turned off :/

KarimAkra:
oh yea this also fixes an issue with songs starting while still loading if they have skipCountdown set to true

https://github.com/user-attachments/assets/5fbc450d-24bd-4437-8f3d-7e43589eaade

pisayesiwsi:
The only issue with this approach is that Haxe Threads cannot be freed, so load a few songs; and you run out of threads to continue doing this same check

ShadowMario:
Threads for this is not an optimal approach.

KarimAkra:
> The only issue with this approach is that Haxe Threads cannot be freed, so load a few songs; and you run out of threads to continue doing this same check

Not really Haxe shutdowns and closes the thread the moment the job function you've passed finishes executing
Which in this case is when we stop the while loop that's running in the thread

You can check that on HaxeThread [here](https://github.com/HaxeFoundation/haxe/blob/development/std%2Fcpp%2F_std%2Fsys%2Fthread%2FThread.hx#L119-L133)

If the thread was still running it would've threw a null object reference when you exit PlayState too since all the fields the loop uses are instance fields

PR #16300, opened 2025-02-13
fix video crash when exiting state

inky03:
simply separates the part that calls the VideoSprite finish callback so it isnt called when the video is just destroyed, which for instance causes a crash in PlayState. This should allow it so you can pause during mid-song cutscenes without risking a game crash due to exiting to the menu

### DIFF
--- a/source/objects/VideoSprite.hx
+++ b/source/objects/VideoSprite.hx
@@ -80,8 +80,10 @@ class VideoSprite extends FlxSpriteGroup {
 			cover.destroy();
 		}
 
-		if(finishCallback != null)
+		if(finishCallback != null) {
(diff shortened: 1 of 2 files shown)


Rom4chek:
this fix crashed for me once it tried to do function "remove(cover)"

inky03:
> this fix crashed for me once it tried to do function "remove(cover)"

just edited a part in the code, but i cant find a way to reproduce it yet. maybe more details would be useful

Rom4chek:
Ok edit you made doesnt make it crash now, yay

KELOTOMU (approved):
sure why not

2JENO (approved):

PR #16302, opened 2025-02-13
support passing Object to lua

inky03:
Object type can be converted to lua so this simply changes a few lines of code to properly support it as a return value in the runHaxeFunction and runHaxeCode functions


KELOTOMU (approved):
gud small change

lunacynne (approved):

PR #16309, opened 2025-02-14
Put some defines around achievements

CrusherNotDrip:
This PR puts Achievements that requires defines like week1-7_nomiss around a define like BASE_GAME_FILES.

I'm also not gonna lie, I didn't even test to see if it works but that would be doubting my code if I need to test it anyways 🤑🤑🤑


TechnikTil (approved):
10/10 would recommend to my dead grandfather

PR #16384, opened 2025-02-28
use power of 10 instead of "for (i in 0...decimals)"?

Justine101213k:


HIGGAMEON:
I hope this gets accepted 😔

PR #16391, opened 2025-03-01
Adding compatibility to V Slice FNF Character Jsons

HIGGAMEON:
This will make V Slice character jsons functionable in Psych Engine.
 settings. I hope this pull helps! If it is not needed, then I understand👍

I used the QT rewired mod as an example!
https://gamebanana.com/mods/566597

image
image


2JENO (approved):

ShadowMario:
I don't see much use for this, I'd rather let people port their work *from* Psych Engine to V-Slice than let people do "Psych ports" even more easier

HIGGAMEON:
Okay that is actually very fair, I didn't think about this being a way to make more Psych ports

PR #16409, opened 2025-03-05
windows dark mode/color support + dpi fix + cpp cleanup

lunacynne:
semi mirror pr of https://github.com/FunkinCrew/Funkin/pull/4267

- moves window functions into `backend.Native`
- better dpi awareness (and draws black for a frame so theres no ugly white border)
- added dark theme support toggle function
- added window bar/text/border color functions
- also prunes some of the existing cpp code that is unused

https://github.com/user-attachments/assets/4374d9d0-adf5-43be-9dd7-ce4316bd987c

### DIFF
--- a/source/Main.hx
+++ b/source/Main.hx
@@ -45,24 +45,10 @@ import backend.Highscore;
 @:cppInclude('./external/gamemode_client.h')
 @:cppFileCode('#define GAMEMODE_AUTO')
 #end
-#if windows
-@:buildXml('
-<target id="haxe">
-	<lib name="wininet.lib" if="windows" />
-	<lib name="dwmapi.lib" if="windows" />
-</target>
-')
-@:cppFileCode('
-#include <windows.h>
-#include <winuser.h>
-#pragma comment(lib, "Shell32.lib")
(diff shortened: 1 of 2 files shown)


FearmongerWade:
thumbs up

2JENO (approved):

DM-kun (approved):

ShadowMario:
This PR has some problems running on Windows 10 and previous.
If possible, don't make it turn dark mode on by default.

ShadowMario:
> This PR has some problems running on Windows 10 and previous. If possible, don't make it turn dark mode on by default.
> 
After further thinking, I might reject this PR because of the Dark mode, it's kinda buggy

cyn0x8:
i can emulate win10 to figure out what might be going on

DMMaster636:
tho, an easy way to fix this is to just redraw the header, so for example enabling and disabling borderless
`
flixel.FlxG.stage.window.borderless = true;
flixel.FlxG.stage.window.borderless = false;
`
it may not be the most optimal solution (because there is a little flash/jitter), but it works

ShadowMario:
Windows 10 has no native support for changing window colors and stuff, so I'm going to have to close this, the PR will be partially implemented and credited though

PR #16471, opened 2025-03-16
Fix double file loading for some reason?

glowsoony:


ShadowMario:
This function is meant to get all files with the same name through all directories available at the moment, wouldn't your change make it unable to load files inside the current level's folder unless forced?

glowsoony:
> This function is meant to get all files with the same name through all directories available at the moment, wouldn't your change make it unable to load files inside the current level's folder unless forced?

Don't see why this would be the case? I've tested this change and have been able to get everything to load correctly. The reason this change is here is because a bug occurs where scripts load twice, with this change that doesn't occur anymore. From what I've tested the engine runs completely fine with this code, unless I have changed something that allows me to load these files?

PR #16542, opened 2025-03-31
new time formatter function

An-enderman:
I added a new time formatter that supports hours, so that longer songs can display the time on the timebar correctly
the added function it's in CoolUtil


ShadowMario:
Pretty useless ngl

